PR #12, opened 2017-12-19
Fix Plugin-Development-Project

d3xter:
Right now without "modules.xml" Intellij IDEA is not able to create a Plugin-Project from the sources.

After providing the ".idea/modules.xml" file, Intellij happily imports the checked out sources as a Plugin-Project


klazuka:
Thanks!

PR #33, opened 2018-04-23
Add Kotlin to the Plugin-Project

d3xter:
This PR adds Kotlin to the eplug-Module, so that when cloning this repo you do not have to setup Kotlin


klazuka:
Thanks, but this has some undesirable consequences. When I prepare the plugin for deployment, previously it produced a 433KB JAR. But with this change it produces a 4+MB ZIP that includes the Kotlin runtime. IntelliJ docs say that you should not bundle the Kotlin runtime with your plugin:
> Please note that you should not include kotlin-runtime and kotlin-stdlib jars with your plugin because Kotlin guarantees backward- and forward- binary compatibility.

([source](https://www.jetbrains.org/intellij/sdk/docs/tutorials/kotlin.html))

What problem did you run into? Because when I tried setting up plugin development on my new laptop, I didn't have any problem with the Kotlin config.

klazuka:
Ok. I haven't spent too much time on the build configuration because I eventually want to switch to defining the build in Gradle (http://www.jetbrains.org/intellij/sdk/docs/tutorials/kotlin.html#kotlin-gradle-plugin).

d3xter:
Alright, finally had some time to fix this.

Now the "eplug"-module has Kotlin configured by default (requires kotlin 1.2),
but the jar-libraries are not included in the generated plugin

PR #44, opened 2018-05-23
Added basic Elm-Live-Templates

d3xter:
This PR adds a couple of Live-Templates:

-  "fn0/1/2/3" to create a Function with 0, 1, 2 or 3 arguments
- "mod" to create a Module Declaration Line
- "ty" to create a Type
- "tya" to create a Type Alias
- "let1/2/3" to create a Let-Expression with 1, 2 or 3 variables
- "case1/2/3" to create a case expression with 1, 2 or 3 branches

All of those comply with the current elm-style-guide.

### DIFF
--- a/src/main/kotlin/org/elm/ide/livetemplates/ElmLiveTemplateContext.kt
+++ b/src/main/kotlin/org/elm/ide/livetemplates/ElmLiveTemplateContext.kt
@@ -1,4 +1,4 @@
-package org.elm.ide
+package org.elm.ide.livetemplates
 
 import com.intellij.codeInsight.template.TemplateContextType
 import com.intellij.psi.PsiFile
--- a/src/main/kotlin/org/elm/ide/livetemplates/ElmLiveTemplateProvider.kt
+++ b/src/main/kotlin/org/elm/ide/livetemplates/ElmLiveTemplateProvider.kt
@@ -1,0 +1,9 @@
+package org.elm.ide.livetemplates
+
+import com.intellij.codeInsight.template.impl.DefaultLiveTemplatesProvider
+
+class ElmLiveTemplateProvider : DefaultLiveTemplatesProvider {
+    override fun getDefaultLiveTemplateFiles() = arrayOf("liveTemplates/Elm")
+
+    override fun getHiddenLiveTemplateFiles() = arrayOf<String>()
+}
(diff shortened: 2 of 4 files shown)


klazuka:
Cool. One thing you might want to check is how this behaves with the stuff I did for ["smart enter"](https://github.com/klazuka/intellij-elm/commit/d4cf5f6fa1da635e272d7622859bce82ab6014f2) a couple weeks ago. Does it interfere? Does it make some things not very useful (i.e. `fn0` might not be worth it now)

What's the idea about the let1/2/3 variants? When I write these things, I rarely know how many local declarations I will need.

When would you use case1? case2 is definitely useful (Maybe, Result, etc). case3 i'm unsure about.

klazuka:
case1 is useful. I've been going through my SPA app and there are a few pages where I have a single value for a `Msg` type, and the `update` function is a `case` with a single branch. 👍 

I'm busy nailing down the Elm 0.19 stuff, but I'll see what I can do about getting this merged in. I want to play with it a little bit first, though.

klazuka:
I just tried this out--it's great! I'm going to merge now, but there are a few things that I think should be fixed in a subsequent PR.

- Restrict context where live template is offered. e.g. the module declaration should never complete `mod` when writing `model` in an update function. Not sure if IntelliJ allows for this.
- type alias should keep the equals on the first line to match `elm-format` style
- why does the type alias live template say "on a single line"? I would get rid of that part.
- let/in inner decl: after typing decl name and inserting the equals sign, should put cursor on the following line, indented, to match `elm-format` style.
- let/in: cursor should not be indented on the line after the `in` (see elm-format 0.7.0/0.8.0). It should be aligned with the `in`.
- module declaration should make the module name customizable because sometimes you need to include the parent directory. e.g. `src/Page/Login.elm` would have a module name of `Page.Login`

Possible new features:
- a template for `if/then/else`
- a template for `case/of` on a Maybe, expanding `Just` and `Nothing`. Maybe even do the same thing for `Result`.

klazuka:
You can restrict the context. See how intellij-rust did it [here](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/resources/org/rust/ide/liveTemplates/output.xml).

Also, it would be good to add [tests like intellij-rust](https://github.com/intellij-rust/intellij-rust/blob/master/src/test/kotlin/org/rust/ide/template/RsLiveTemplatesTest.kt) for Live Template expansion. We don't need to test each and every live template, but we should make sure that at least one expands (proving that the path to the XML file is correct) and that any live templates which have a restricted context are working correctly.

d3xter:
I've included "let1/2/3" to be consistent with case/fn/etc.

I'll have a look the proposed fixes and implement them when I have time.

About the context:
Right now they are all limited to Elm-files as a whole, 
will take a look how intellij-rust restricts them to specific parts in an File


About the template for `case Maybe of`:
We could do that at first, later an Intention to expand a `case of` 
to create a branch for every/all missing Type-Constructors would be nice.

PR #45, opened 2018-05-26
Add Elm 0.19 support, drop 0.18

klazuka:
Most of the Elm 0.19 changes were easy, but the most difficult thing to fix was the moving of package dependencies out of the local `elm-stuff` directory and into `~/.elm`. Prior to this, we could cheat and include `elm-stuff` within the source root, making PsiReference resolve just work. But now that all projects' dependencies are going in the same place (and with different versions), we needed to parse the project dependencies in `elm.json` in order to create proper library source roots for each dependency.

I'm merging this in to `master` now so that all future work is based off of it. Obviously I will not ship this until Elm 0.19 is officially released. At which point I will also provide a link to an old JAR which is compatible with Elm 0.18.


klazuka:
@d3xter FYI. There's still a lot of stuff in the project internals (`ElmWorkspaceService`) which I need to finish. But this should be good enough to do some stuff related to run configs (I seem to remember you saying something about doing an `elm-reactor` run config... that would be cool, same goes for repl).

PR #50, opened 2018-07-20
Add quick documentation support for some types

ajalt:
This PR add preliminary support for documentation generation. This PR is limited in scope to `ElmFunctionDeclarationLeft` elements (i.e. simple functions), and union types. Support for more element types can be added in the future.

In order to implement this, I needed to fix the comment lexing to generate a single token for each block comment, rather than the arbitrary number of tokens that are currently generated.

Let me know what you think about the current output format, and what additional elements you'd like to see supported.


ajalt:
Here are some examples:

example1

example2

ajalt:
Added support for union types:
example3

klazuka:
Thanks!

I'm so glad that you simplified the lexing of block comments.

The only bug I found is that `Result.andThen` does not find the doc comment. It does get the type annotation though. I traced it in the debugger, and it seems to be a problem with skipping over the whitespace and "virtual end decls" to find the block comment.

I'm going to merge this as-is, and we can refine it later.

PR #51, opened 2018-07-22
Insert virtual end tokens earlier

ajalt:
Here's a potential solution to the problem raised in #50.

The actual issue was not in `ElmDocumentationProvider`, but is due to the way virtual tokens are inserted into the stream. Currently, virtual end tokens are inserted directly before the token that decreases the indentation. This causes any comments between the closed expression and the virtual token to be parsed as part of the expression, rather than top level comments.

For example, in a case statement:

capture

This PR addresses the issue by inserting the virtual end token as close to the closed expression as possible. This allows the structure to be parsed correctly:

capture2

However, the implementation is rather fragile due to the way the virtual tokens are combined with real tokens during parsing. For reasons I don't fully understand, end decl tokens are combined with the immediately preceding token into a `PsiElement` with a type of `VIRTUAL_END_DECL`, but the text of the preceding token, and the preceding token is not present in the PSI.

To work around this behavior, I inserted the end tokens after the first whitespace token after the last code token in the closed expression.


src/test/kotlin/org/elm/ide/typing/ElmOnEnterSmartIndentHandlerTest.kt
```diff
@@ -65,13 +65,13 @@ f =
     fun `test after equals in let declaration partial`() = doTest("""
 f =
     let
-        x ={-caret-}
+        x = {-caret-}
```
ajalt: So this test was actually broken. The problem seems to be that the end decl token from the `in` now comsumes the newline that is generated when you type `\n`, so IntelliJ thinks you're lined up with the `let`, rather than the `x`. I don't know of a good way to fix this without adding extra whitespace, which I did here.

klazuka: I have fixed this on my side. The problem was in `ElmOnEnterSmartIndentHandler`. I'm going to merge your PR and then apply my fix.

klazuka:
This is a nice improvement. Thank you.

PR #55, opened 2018-08-02
Add clickable links to generated documentation

ajalt:
untitled

This PR adds reference links to the generated documentation. Clicking one will show the documentation for the linked element.

This makes the generation a little more involved, but it's still pretty straight forward. 

For qualified names (like `Http.Status`), I decided to make the entire id a single link to the referenced element, rather than making the qualification sections link to modules.

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/psi/elements/ElmTypeAliasDeclaration.kt
+++ b/src/main/kotlin/org/elm/lang/core/psi/elements/ElmTypeAliasDeclaration.kt
@@ -1,13 +1,13 @@
 package org.elm.lang.core.psi.elements
 
 import com.intellij.lang.ASTNode
+import com.intellij.psi.PsiComment
 import com.intellij.psi.PsiElement
 import com.intellij.psi.stubs.IStubElementType
 import com.intellij.psi.util.PsiTreeUtil
 import org.elm.ide.icons.ElmIcons
-import org.elm.lang.core.psi.ElmStubbedNamedElementImpl
+import org.elm.lang.core.psi.*
 import org.elm.lang.core.psi.ElmTypes.UPPER_CASE_IDENTIFIER
-import org.elm.lang.core.psi.IdentifierCase
 import org.elm.lang.core.stubs.ElmTypeAliasDeclarationStub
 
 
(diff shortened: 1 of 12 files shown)


src/main/kotlin/org/elm/lang/core/psi/elements/ElmTypeAliasDeclaration.kt
```diff
@@ -47,12 +47,7 @@ class ElmTypeAliasDeclaration : ElmStubbedNamedElementImpl<ElmTypeAliasDeclarati
...
-    // TODO [kl] this will be wrong in the case of a function type ref
-    // such as `Int -> { foo: String }`. We want to know that it is exclusively a record.
-    // We should see if we can get GrammarKit to parse [ElmTypeRef] into a better data structure
-    // than what Kamil had.
+    /** `true` if the alias is exclusively a record */
     val isRecordAlias: Boolean
-        get() = getStub()?.isRecordAlias
-                ?: typeRef?.recordTypeList?.isNotEmpty()
-                ?: false
+        get() = stub?.isRecordAlias ?: (typeRef?.firstChild is ElmRecordType)
```
ajalt: This works because the type ref class wraps the record manually, so there's no way for there to be whitespace etc. before it.

src/test/kotlin/org/elm/ide/docs/ElmResolveLinkTest.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    fun `test type alias`() = doTest(
+            """
+type alias Foo = Int
+         --X
+
+foo : Foo
+foo = 0
+--^
+""", "Foo")
```
ajalt: I wanted to add a test like this:

```kotlin
    fun `test qualified type`() = doTest(
            """
--@ Foo.elm
module Foo exposing (Bar)

type Bar = Bar
   --X

--@ main.elm
import Foo

foo : Foo.Bar
foo = 0
--^
""", "Foo.Bar")
```

But I couldn't get the `fileTreeFromText` process working with a marker in two different files. Let me know if you have an idea on how to set this up.

ajalt:
I added documentation for type aliases and function parameters, and module declarations, and added union members to type declaration docs. I believe this is all the remaining types that should get doc support. Let me know if I missed any.

Please suggest any improvements to the format of the generated docs. I looked into adding the type to function parameters when the function is annotated, but walking through all the different ways you can do pattern matching is complex enough that I'll defer that for now.
alias
module
parameter
union

klazuka:
- type aliases and function parameters, and module declarations, and added union members to type declaration docs

PR #57, opened 2018-08-03
Add more module information to rendered docs

ajalt:
This PR addresses a comment in #55 by adding the module that elements are defined in to their docs.

defined

It also adds rendered doc comments for modules. Modules can have `@doc` directives that are defined to inline the named elements into the rendered docs. I decided to just render links to the elements rather than inlining the docs, in order to make the result more manageable.

module

### DIFF
--- a/src/main/kotlin/org/elm/ide/docs/ElmDocumentationProvider.kt
+++ b/src/main/kotlin/org/elm/ide/docs/ElmDocumentationProvider.kt
@@ -6,8 +6,11 @@ import com.intellij.lang.documentation.DocumentationMarkup
 import com.intellij.openapi.util.text.StringUtil
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiManager
-import org.elm.lang.core.psi.*
+import org.elm.lang.core.psi.ElmDocTarget
+import org.elm.lang.core.psi.ElmPsiElement
 import org.elm.lang.core.psi.ElmTypes.BLOCK_COMMENT
+import org.elm.lang.core.psi.ancestors
+import org.elm.lang.core.psi.elementType
 import org.elm.lang.core.psi.elements.*
 import org.elm.lang.core.resolve.scope.ImportScope
 import org.elm.lang.core.resolve.scope.ModuleScope
@@ -157,7 +160,7 @@ private fun documentationFor(decl: ElmModuleDeclaration): String? = buildString 
     }
 
     renderDocContent(decl) { html ->
-        val declarations = ModuleScope(decl.elmFile).getDeclaredValues()
+        val declarations = ModuleScope(decl.elmFile).run { getDeclaredValues() + getDeclaredTypes() }
 
         // Render @docs commands
         html.replace(Regex("<p>@docs (.+?)</p>", RegexOption.DOT_MATCHES_ALL)) { match ->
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/elm/ide/docs/ElmDocumentationProvider.kt
```diff
@@ -157,6 +159,17 @@ private fun documentationFor(clause: ElmAsClause): String? = buildString {
     return documentationFor(decl)
 }
 
+private fun StringBuilder.renderDefinitionLocation(element: ElmPsiElement) {
+    val module = element.elmFile.getModuleDecl() ?: return
+    renderDefinedInQIDs(module.upperCaseQID.upperCaseIdentifierList)
+}
+
+private fun StringBuilder.renderDefinedInQIDs(ids: List<PsiElement>) {
```
klazuka: Assuming the comment on 153 is applied, then instead of using `joinTo()` to rebuild the dotted string, I'd just use `getText()` on the QID itself.

`getText()` should also be used on line 174 (`ref.upperCaseQID.upperCaseIdentifierList.joinToString(".") { it.text }`) rather than rebuilding the dotted string.

src/main/kotlin/org/elm/ide/docs/ElmDocumentationProvider.kt
```diff
@@ -144,11 +150,7 @@ private fun documentationFor(decl: ElmModuleDeclaration): String? = buildString
     definition {
         i { append("module") }
         append(" ", ids.last().text)
-
-        if (ids.size > 1) {
-            i { append(" declared in ") }
-            ids.dropLast(1).joinTo(this, ".") { it.text }
-        }
+        renderDefinedInQIDs(ids.dropLast(1))
```
klazuka: Rather than display the nested module `Foo.Bar` as `Bar defined in Foo`, I would rather just show it as `Foo.Bar` and get rid of the defined-in clause. I never think about sub-modules independently of the parent module.

And then `renderDefinedInQIDs` can either be inlined or changed so that it takes an upper-case QID parameter instead of a list of psi elements.

src/main/kotlin/org/elm/ide/docs/ElmDocumentationProvider.kt
```diff
@@ -152,6 +152,19 @@ private fun documentationFor(decl: ElmModuleDeclaration): String? = buildString
         append(" ", ids.last().text)
         renderDefinedInQIDs(ids.dropLast(1))
     }
+
+    renderDocContent(decl) { html ->
+        val declarations = ModuleScope(decl.elmFile).getDeclaredValues()
+
+        // Render @docs commands
+        html.replace(Regex("<p>@docs (.+?)</p>")) { match ->
```
klazuka: Elegant solution 💯. I was expecting that doing something like this would require lexing the contents of block comments and creating PsiElements, but this is a great, non-invasive solution for right now.

Eventually I _do_ want to parse it so that we can have references from identifiers in doc comments to identifiers in source code. Right now the rename refactoring is blind to these "soft" linkages.

src/test/kotlin/org/elm/ide/docs/ElmQuickDocumentationTest.kt
```diff
@@ -239,6 +239,29 @@ main = ()
...
+    // the module element, we need to put the line comment inside the doc comment.
+    fun `test module with docstring`() = doTest(
+            """
+module Main exposing (main)
+{-|  --^
+
+Module docs
+
+# Header
+@docs main
```
klazuka: When I was running your new code, I noticed that the module documentation's embedded links are broken when the list of identifiers in a section stretches across multiple lines. Here's an [example from Html.Events](https://github.com/elm/html/blob/master/src/Html/Events.elm#L20). 

screen shot 2018-08-03 at 9 49 01 pm

PR #63, opened 2018-08-09
Add interface tags to help organize PSI

ajalt:
Many PSI elements have a list of children that can be one of several types. Currently, it's onerous to iterate these children. This PR adds new interfaces to help group elements. These interfaces have no methods; they merely act like tags. 

Unfortunately, Kotlin doesn't have union types, and sealed classes can't be interfaces, so this doesn't allow the use of exhaustive `when`. Still, it does make working with the PSI simpler.


klazuka:
@ajalt my initial reservation is the extra cognitive load of introducing more names into the system. I'm not that smart, and it's hard for me to keep this all in my head (it doesn't help that the existing names that I came up with are not that good). So let's just add a `Tag` suffix to each of the interface names.

For reference, the Elm compiler's first-stage AST is here: https://github.com/elm/compiler/blob/master/compiler/src/AST/Source.hs

ajalt:
I updated the BNF to generate a few new element types. Now none of the tags are lies: if an element has children, you can iterate over all of the non-structural children in order, and they'll all have the same tag.

klazuka:
Nice.

I wonder if there's a better way to represent the literals. The Kotlin plugin does the following:
screen shot 2018-08-11 at 10 17 59 am

Their `VALUE_ARGUMENT` node appears to wrap all possible things that could be passed in the argument list to a function. And they have an `INTEGER_CONSTANT` element to wrap the underlying number token. But that just seems to move things up a level, and I don't know what the tradeoffs are. 

I'm going to merge this PR as-is since it's an improvement over the status quo. Thanks!

PR #67, opened 2018-08-23
Combine InnerValueDeclaration with ValueDeclaration

ajalt:
The code for value declarations and type annotations was duplicated between top-level and nested occurrences. This PR uses the same element class for both.

It also fixes a bug where inner type annotations weren't reference elements, so wouldn't get picked up by rename etc.


src/main/kotlin/org/elm/lang/core/psi/elements/ElmTypeAnnotation.kt
```diff
@@ -54,6 +57,7 @@ class ElmTypeAnnotation(node: ASTNode) : ElmPsiElementImpl(node), ElmReferenceEl
     override val referenceName: String
         get() = referenceNameElement.text
 
-    override fun getReference() =
-            LocalTopLevelValueReference(this)
+    override fun getReference(): ElmReference =
+            if (parent is PsiFile) LocalTopLevelValueReference(this)
```
ajalt: Do I need to do anything different here to account for stubs?

klazuka: Nope, this is fine as-is. The Psi stubs only come into play for declarations that (1) introduce a name (e.g. a `PsiNamedElement`) and (2) are visible outside of the file in which it is defined (they can be imported)

PR #70, opened 2018-08-27
Add code folding

ajalt:
This PR adds code folding to all regions that I thought made sense. It follows the global settings for default collapse, so depending on your settings, module declarations, imports, and doc comments might be collapsed when you open a file.

capture

### DIFF
--- a/src/main/kotlin/org/elm/ide/folding/ElmFoldingBuilder.kt
+++ b/src/main/kotlin/org/elm/ide/folding/ElmFoldingBuilder.kt
@@ -1,0 +1,117 @@
+package org.elm.ide.folding
+
+import com.intellij.codeInsight.folding.CodeFoldingSettings
+import com.intellij.lang.ASTNode
+import com.intellij.lang.folding.FoldingBuilderEx
+import com.intellij.lang.folding.FoldingDescriptor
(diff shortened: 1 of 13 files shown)


klazuka:
This is great: simple, clean implementation. 

I do have a few minor nits:

- when I ran it in the IDE, imports were not collapsed by default (even though in Settings -> Editor -> General -> Code Folding the checkbox for "Imports" is checked). Maybe it's just a problem with running in the sandbox—my settings are always hosed.
- when collapsing imports, it shows as `...`, but Kotlin folding displays their collapsed imports as `import ...`, which is nicer.
- no tests. Looks pretty easy to add: see [IntelliJ docs](http://www.jetbrains.org/intellij/sdk/docs/tutorials/writing_tests_for_plugins/folding_test.html?search=testing) and intellij-rust example [RsFoldingBuilderTest](https://github.com/intellij-rust/intellij-rust/blob/master/src/test/kotlin/org/rust/ide/folding/RsFoldingBuilderTest.kt).

Thank you.

ajalt:
Imports are collapsed by default, but if you navigate to the definition of a module (rather than something defined in the module), your cursor will move inside the module PSI element being collapsed, causing it to expand. I'm not sure if there's a way around that.

PR #71, opened 2018-08-28
Enable filtering out imports in the Find Usages panel.

ajalt:
I'd also like to be able to filter out type annotations when finding usages of a function, but there's no button for that. Maybe we could use the "show write access" or "show read access" button? It would be kind of gross, but they aren't currently hooked up, and they aren't really meaningful in a pure language.

filtered
filtered2


klazuka:
I didn't know that this extension point existed. Nice! I was also annoyed by imports showing up in Find Usages.

Since importing an Elm module has no side-effects, it seems totally reasonable to ignore them. If people ask for it, we can always add a user setting to show them.

PR #74, opened 2018-08-31
Update dependencies

ajalt:


build.gradle
```diff
@@ -107,11 +103,6 @@ apply plugin: 'org.jetbrains.grammarkit'
 import org.jetbrains.grammarkit.tasks.GenerateLexer
 import org.jetbrains.grammarkit.tasks.GenerateParser
 
-grammarKit {
```
klazuka: Why is it safe to remove these? Are they implicit due to the line `classpath "com.github.JetBrains:gradle-grammar-kit-plugin:2018.1.7"` above? I’m happy to get rid of them—just curious.

PR #82, opened 2018-09-07
Support expression accessors

ajalt:
This PR adds support for a previously overlooked bit of syntax: record field accessors on record literals and parenthesized expressions.

For example:

```elm
(foo).bar.baz

{x={y=2}}.x.y
```

This PR adds two new PSI rules (`ExpressionWithAccessor` and `RecordWithAccessor`), to contain these cases without affecting other rules. The accessor is parsed with yet another manual parser, `ExpressionAccessorParser`.


ajalt:
The manual dot parser was an excellent idea.

Issue #19, opened 2018-02-24
Some multiline string literals can cause a crash


muelli:
The plugin seems to have crashed:

```
java.util.NoSuchElementException
	at java.util.LinkedList.removeFirst(LinkedList.java:270)
	at java.util.LinkedList.pop(LinkedList.java:801)
	at org.elm.lang.core.lexer.ElmLayoutLexer.advance(ElmLayoutLexer.kt:160)
	at com.intellij.lang.impl.TokenSequence$Builder.performLexing(TokenSequence.java:86)
	at com.intellij.lang.impl.PsiBuilderImpl.a(PsiBuilderImpl.java:202)
	at com.intellij.lang.impl.PsiBuilderImpl.<init>(PsiBuilderImpl.java:175)
	at com.intellij.lang.impl.PsiBuilderImpl.<init>(PsiBuilderImpl.java:130)
	at com.intellij.lang.impl.PsiBuilderFactoryImpl.createBuilder(PsiBuilderFactoryImpl.java:52)
	at com.intellij.psi.tree.ILazyParseableElementType.doParseContents(ILazyParseableElementType.java:86)
	at com.intellij.psi.tree.IFileElementType.parseContents(IFileElementType.java:49)
	at com.intellij.psi.impl.source.tree.LazyParseableElement.ensureParsed(LazyParseableElement.java:192)
	at com.intellij.psi.impl.source.tree.LazyParseableElement.getLastChildNode(LazyParseableElement.java:246)
	at com.intellij.psi.impl.source.tree.LazyParseableElement.getLastChildNode(LazyParseableElement.java:39)
	at com.intellij.psi.stubs.DefaultStubBuilder$StubBuildingWalkingVisitor.a(DefaultStubBuilder.java:123)
	at com.intellij.psi.stubs.DefaultStubBuilder$StubBuildingWalkingVisitor.visitNode(DefaultStubBuilder.java:93)
	at com.intellij.psi.stubs.DefaultStubBuilder$StubBuildingWalkingVisitor.buildStubTree(DefaultStubBuilder.java:83)
	at com.intellij.psi.stubs.DefaultStubBuilder.buildStubTreeFor(DefaultStubBuilder.java:58)
	at com.intellij.psi.stubs.DefaultStubBuilder.buildStubTree(DefaultStubBuilder.java:39)
	at com.intellij.psi.stubs.StubTreeBuilder.buildStubTree(StubTreeBuilder.java:84)
[...]

PR #84, opened 2018-09-07
Improve parsing of string literals

### LINKED ISSUE #78: Lexer/highlighter gets confused with string literals in some cases
The way the plugin handles strings often leads to error highlights when typing a string with a symbol in it:

```
helloWorld =
    "Hello, world!"
    ^             ^
```

This only seems to happen to the last string in a file (presumably because the string pattern matches across lines and uses the next `"` as the closing quote). Inserting any character before the first symbol resolves the highlighting for the string. There isn't actually any errors showing on the right bar, just error highlighting on the text (without a tooltip).

I think ideally quotes would be handled as for e.g. Java:

```
helloWorld =
    |
```

On typing a quote `"`, a quote is inserted at the cursor's position and one immediately after it.

```
helloWorld =
    "|"
```

If you type a `"` while the cursor is immediately to the left of another quote, the cursor is moved to the right of the `"` rather than inserting an additional quote.

```
helloWorld =
    ""|
```

ajalt:
The lexer currently requires the closing quote for string literals. So while typing, the string is not highlighted. In some cases, the string will not be highlighted correctly even after typing the closing quote. This happens when the lexer emits tokens after the string error token. In that case, the partial lex doesn't look at the start of the string again, so more error tokens are emitted.

This PR changes the lexer to make closing quotes in string literals optional. An annotator was added to report the missing closing quotes, although since all strings literals in elm are multi-line, the error is always reported at the end of the file.

Fixes #78


klazuka:
I looked at the Kotlin Psi and they lex the string literal quote delimiter as its own token:

screen shot 2018-09-06 at 7 56 55 pm

I'd be inclined to handle it similarly where we parse the string literal from the constituent tokens. The advantage is that it would not require the error annotator, and it would put us in good position to highlight escape sequences. e.g. the `\n` in the following Kotlin string:

screen shot 2018-09-06 at 8 04 04 pm

Thoughts?

ajalt:
Ok, I split strings into individual tokens, which allows highlighting of escape characters, including error highlights on invalid escapes.

Since I had to create a new `StringConstant` rule anyway, I made rules for the other two literal types as well, so we no longer need one multi-purpose `ElmLiteral` element.

Here's all the available highlighting:

capture

By the way, in typical Elm fashion, no where is it documented what the valid escape sequences are. But if you compile a program with an invalid escape, the error message includes the complete list.

klazuka:
Nice! And it fixes #19.

I noticed while editing a program that a partial string literal would still result in cascading parse errors further down in the file (which is no worse than the status quo which behaves the same way). In the notes for this PR you said `all strings literals in elm are multi-line`, but I tested it out a normal string literal containing a newline and got the following error in Ellie:

screen shot 2018-09-07 at 7 56 57 pm

How hard would it be to disallow newlines in a regular string literal?

ajalt:
I don't know what happened to by brain: I was positive that you could have newlines in single-quotes strings.

Anyway, now you get a big red error if you forget the closing quote:

capture

src/test/resources/org/elm/lang/core/parser/fixtures/partial/Strings.txt
```diff
@@ -0,0 +1,59 @@
...
+    ElmExpression(EXPRESSION)
+      ElmCharConstant(CHAR_CONSTANT)
+        PsiElement(CHAR_LITERAL)(''d')
+  PsiElement(VIRTUAL_END_DECL)('\n')
+  PsiWhiteSpace('\n')
+  ElmValueDeclaration(VALUE_DECLARATION)
+    ElmFunctionDeclarationLeft(FUNCTION_DECLARATION_LEFT)
+      PsiElement(LOWER_CASE_IDENTIFIER)('c')
+    PsiElement(EQ)('=')
+  PsiElement(OPEN_QUOTE)('"')
```
ajalt: I was hoping to be able to still parse these as a string constants, rather than just unexpected tokens, but no amount of fiddling with `pin` and `recoverWhile` got that to work. Do you of know a way to do it?

klazuka:
GrammarKit is a pain. It took me at least an hour, but I have something that works. I pushed a branch consisting of a single commit https://github.com/klazuka/intellij-elm/commit/8bd920558557931cd3e4053348cf0fa151f129f3 which you can apply to your branch.

This [message board post](https://intellij-support.jetbrains.com/hc/en-us/community/posts/360000081950/comments/360000071084) about GrammarKit `pin` and `recoverWhile` is pretty useful. I ended up doing what he recommended: decomposing the parse rule into an outer and an inner, putting the `pin` on the outer and the `recoverWhile` on the inner. 

The really nasty thing about `recoverWhile` is that it runs both in the failure _and_ the success case. So if you aren't careful, it will run wild and consume way more tokens than you expect, __even when there is no parse error__. That's why it's helpful to separate the parse rules as discussed in that msg board post above. One of the `recoverWhile` predicates should be the normal end delimiter (e.g. an end quote) in order to handle the success case.

The last tricky bit was figuring out the predicate for the `recoverWhile`. At first I tried introducing a manual/external parse rule to detect newlines, but that didn't work. The thinking was that we just swallow tokens until we reach newline and then hope that the next line starts with something that we can resume from. I settled on hard-coding a few tokens which naturally delimit (or start) a new expression. GrammarKit will eat tokens up until it reaches that delimiter. The delimiter (e.g. a comma) will be left on the token stream, allowing another parse rule to match on it (e.g. if we were parsing a list of comma-separate string literals).

ajalt:
Wow, nice work. Thanks! That post was extremely helpful. The fact that `recoverWhile` runs in the success case explains all the crazy behavior I ran into. What a strange design decision.

jorgengranseth:
Great work both of you! 👏 Will be amazing not to have this annoyance any longer.

klazuka:
@jorgengranseth I just published [version 1.3.2](https://github.com/klazuka/intellij-elm/releases/tag/v1.3.2) on Github. Please give it a try and let me know if you run into any problems. Once it seems stable enough, I will upload it to JetBrains for distribution.

jorgengranseth:
@klazuka looks great! A few observations:

Using a newline in a single-quoted string now gives an error on the right-hand-side bar but no error highlight in the file. The tool-tip I get on the right-hand side says `INVALID_STRING_ESCAPE, REGULAR_STRING_PART or STRING_ESCAPE expected, got ''`. Should possibly say that [strings cannot contain newlines](https://github.com/klazuka/intellij-elm/pull/84#issuecomment-419607879), or does the plugin just offer lexer error messages at this point (can't remember, but seems familiar)? I understand that `INVALID_STRING_ESCAPE` is now a string value in the parser (from this PR), but it may seem strange to a user who wants to fix the error that inserting an invalid string escape will do the job.

Which presumably also causes this: Inserting an `INVALID_STRING_ESCAPE` in a string (e.g. `"Hi\x"`) gives error highlighting in the file but not a tool-tip or an error (similar to how string delimiters were erroneously highlighted in #78, except this time there's an actual error).

Neither of these are serious issues as they're still strict improvements on 1.3.1, but probably worth noting.

Folding seems amazing. 🎉

klazuka:
Agreed, the error reporting is not great. Looks like @ajalt is working on another PR that might be related.

ajalt:
If we want to improve the error reporting, we can combine the two escape tokens and use an annotator to give more helpful error messages. I'll make an issue to track it.

PR #85, opened 2018-09-09
Add quote matching and char escape highlighting

### LINKED ISSUE #79: Smarter insertion of double-quotes
@jorgengranseth wrote in a separate issue:

I think ideally quotes would be handled as for e.g. Java:

```
helloWorld =
    |
```

On typing a quote `"`, a quote is inserted at the cursor's position and one immediately after it.

```
helloWorld =
    "|"
```

If you type a `"` while the cursor is immediately to the left of another quote, the cursor is moved to the right of the `"` rather than inserting an additional quote.

```
helloWorld =
    ""|
```

ajalt:
This PR adds two features:

### Quote matching 

This includes all the expected functionality for string, raw string, and char literals. 

The hooks for this functionality were even more undocumented than usual. I had to dig through the intellij source code to figure this out, so I added comments to help explain what each one does.

Fixes #79 

### Char literal parsing. 

This was helpful in implementing quote matching, with the added bonus that we get all the highlighting that was added to strings in #84.

I chose to implement char parsing the same way we implemented string parsing. I don't know of any other plugin that does this. Instead, they use very permissive regexes in the lexer, the way that we currently do. In order for quote matching to work properly for char literals, the regex has to be so permissive as to accept a single `'` as a valid token. This necessitates adding an annotator to validate char literals. It also prevents highlighting for escape sequences.

Splitting up char literals in the lexer does bring a small complication in that we don't want to emit more than one `REGULAR_STRING_PART` token for each literal, but keeping track of that is simple. Another option would be to have char literals chomp to the end of line like string do, then add an annotator for the error.


src/main/kotlin/org/elm/ide/typing/ElmQuoteHandler.kt
```diff
@@ -0,0 +1,63 @@
+package org.elm.ide.typing
+
+import com.intellij.codeInsight.editorActions.MultiCharQuoteHandler
+import com.intellij.codeInsight.editorActions.QuoteHandler
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.editor.highlighter.HighlighterIterator
+import org.elm.lang.core.psi.ElmTypes.*
+
+class ElmQuoteHandler : QuoteHandler, MultiCharQuoteHandler {
+    // If this returns true, the editor will move the carat one character right instead of inserting the typed character
```
klazuka: I'm glad you wrote all these comments. It's not obvious at all what's going on just based on the method names.

klazuka:
The quote handler stuff is a big usability improvement. Nice work.

PR #89, opened 2018-09-19
Allow FieldAccess as target for function calls

ajalt:
We were incorrectly rejecting syntax like the following:

```elm
input =  { x = \() -> () }
valueRef = { foo = input.x (), bar = () }
                   -------
```

The issue was that `FieldAccess` wasn't an option in the `FunctionCallTarget` rule.

In order to get this to work, I cleaned up the hack that required `FieldAccess` be a higher priority than `FunctionCall`. Now we do a small lookahead in all the whitespace-sensitive rules to make sure they don't match on part of a dotted id.


src/main/kotlin/org/elm/lang/core/psi/ElementTags.kt
```diff
@@ -11,10 +12,12 @@ interface ElmExpressionPartTag : ElmPsiElement
 /** An element that can occur in an [ElmExpression] as the argument to a function or operator */
 interface ElmOperandTag : ElmExpressionPartTag
 
-/** An element that can be the parameter of an [ElmFunctionDeclarationLeft], [ElmAnonymousFunction],
- * or [ElmCaseOfBranch] */
+/** An element that can be the parameter of an [ElmFunctionDeclarationLeft], [ElmAnonymousFunction], or [ElmCaseOfBranch] */
 interface ElmNameDeclarationPatternTag : ElmNamedElement
 
+/** A function being called as the child of a [ElmFunctionCall] */
+interface ElmFunctionCallTarget : ElmReferenceElement
```
klazuka: I don't see any classes getting tagged with this new interface.

src/main/kotlin/org/elm/lang/core/parser/manual/DotWithoutLeadingWhitespaceParser.kt
```diff
@@ -0,0 +1,24 @@
+package org.elm.lang.core.parser.manual
+
+import com.intellij.lang.PsiBuilder
+import com.intellij.lang.parser.GeneratedParserUtilBase
+import com.intellij.lang.parser.GeneratedParserUtilBase.*
+import com.intellij.psi.TokenType
+import org.elm.lang.core.psi.ElmTypes.*
+
+/**
+ * Parses a `.` token iif it does not have whitespace following it.
```
klazuka: s/following/before

src/main/kotlin/org/elm/lang/core/psi/ElementTags.kt
ajalt: good catch, sorry about that

PR #91, opened 2018-09-21
Add all remaining function call targets to grammar

### LINKED ISSUE #90: using << to compose functions creates irrelevant error
Hello,

Thank you for the amazing plugin!

`(Url.percentEncode << String.replace "\"" "\\\"") query` <- error here
generates error but should be ok.

ajalt:
I went through the `Operand` rule and checked all of them to see which ones could ever be valid call targets. I should have done that initially.

Fixes #90

Instead of having an equivalent to `FunctionCallTarget`, the Elm compiler allows any operand as a call target, and issues errors from the type inference stage. I'd like to do the same, but defining `FunctionCall ::= Operand Operand+` caused empty records to parse incorrectly, for a reason I couldn't identify.

Additionally, I noticed that, unlike Kotlin and many other languages, Elm forbids whitespace after the `-` tokens in negation expressions. `x = - 1` is a syntax error. I added a manual parser to match this behavior.


klazuka:
Thanks. I’ll look at it tonight and hopefully submit a new build to JetBrains.

src/main/kotlin/org/elm/lang/core/psi/elements/ElmFunctionCall.kt
```diff
@@ -12,11 +15,13 @@ import org.elm.lang.core.psi.*
...
+ *  - `(\x -> x) 1`
+ *  - `(a << b) c
  */
 class ElmFunctionCall(node: ASTNode) : ElmPsiElementImpl(node), ElmOperandTag {
     /** The function or operator being called */
     val target: ElmFunctionCallTarget get() = findNotNullChildByClass(ElmFunctionCallTarget::class.java)
 
     /** The arguments to the function. This will always have at least one element */
-    val operands: Sequence<ElmOperandTag> get() = directChildren.filterIsInstance<ElmOperandTag>()
+    val operands: Sequence<ElmOperandTag> get() = directChildren.filterIsInstance<ElmOperandTag>().drop(1)
```
klazuka: good catch

PR #98, opened 2018-10-18
Fix references to Cmd and Sub

ajalt:
This fixes a regression in #92 that caused implicit references to `Cmd` and `Sub` not to resolve. These two values are special-cased in `moduleLookupHack` when creating the module definition, but not the dependency exposed modules, so `ElmProject.exposes` was returning false because `exposedModules` contained `Platform.Cmd`, but the module declaration name was just `Cmd`

I added the hack to the `exposedModules`, but I don't know if that's the correct fix. Maybe the hack is no longer necessary?


klazuka:
The way I was handling the `Platform.Cmd` alias was lame, lame, lame. The good news is that I've removed it in the [kl/better-18-support branch](https://github.com/klazuka/intellij-elm/tree/kl/better-18-support) (along with a lot of other improvements).

PR #99, opened 2018-10-20
Add package resolution for Elm 0.18 projects; improve the foundation

klazuka:
This PR brings full package resolution to Elm 0.18 projects so that it works similarly to how Elm 0.19 projects already work. This means that we now use your project's source directories and dependencies to figure out which modules are in-scope for purposes of imports/code-completion/etc.

As part of doing this, I also removed some hacks and, in general, tried to make things better:
- the `elm-stuff` directory is now an excluded source root. No more `.elmi` and `.elmo` files junking up your search results!
- I improved the way that `elm/core` is detected when providing implicit imports
- I improved the way that the implicit importing and aliasing of `Platform.Cmd` and `Platform.Cmd` works
- the integration tests can now define Elm projects which download and install Elm package dependencies


klazuka:
@ajalt FYI this was a pretty big change. Hopefully it doesn't cause too many merge conflicts for you.

PR #100, opened 2018-10-21
Add type inference

ajalt:
This PR adds type inference for all psi elements. This information is used in the documentation and expression type providers, and a new inspection that reports errors discovered during inference.

inspection
untitled

I have not yet added unification, so no diagnostics are reported for expressions that involve variable types (e.g. parameters in unannotated functions or parameters annotated with type variables). I have not implemented general pattern coverage, so missing branches in case expressions won't generate a diagnostic.

My approach to these diagnostics is that false positives (diagnostics reported by this plugin but not the elm compiler) are bugs, while false negatives are feature requests that can implemented later.

Type inference inherently requires walking over the entire PSI and handling every possible corner case. This is an unavoidably messy problem to solve, compounded by the fact that compiler plugins have to operate on the PSI. The Elm compiler has the luxury of only needing to run on syntactically correct programs, and on an AST that is several levels of IR removed from the output of the parser. I made an effort to keep the code straightforward, and to comment the subtle parts. But I've been deep in the weeds on this for long enough that I'm sure there are opportunities for improving the maintainability, so please let me know if you think of anything, or if there's any code that's confusing and needs more clarification.

The code is structured so that a single inspection, the `ElmTypeInferenceInspection`, triggers inference for all top level declarations. Inspections only run on open files, so we don't need to do whole-program inference. The inference starts in `TypeInference.kt`, primarily in the `InferenceScope` class. This class walks down a PSI for a declaration and infers the type for each element that has one. Types are represented by the sealed class `Ty`, which has subclasses that match the Elm compilers `Type` class.

When inferring the type of a reference to a declaration, the type annotation is used if there is one, rather than inferring the function body. This is a significant performance improvement: without it, we would potentially have to do whole-program inference for every change to the PSI (including all dependencies). We still use inference for the types on unannotated functions, but given that Elm encourages apps to annotate function, and requires packages to do so, we rarely need to infer more than one declaration at a time. Even if we wanted to ignore annotations and do whole-program inference, a lot of the standard library is implemented in `Elm.Kernel`, and there's no way to infer that synthetic module.

The results of inference are cached on the top level declaration element's user data. Changes within a function body will invalidate that function's cache. Other changes will invalidate all cached inferences.

This is definitely a large PR. Most of the code in this PR is in new files, but I did make some small changes to the BNF and element tags to fix some bugs or make the PSI easier to work with. If you'd like to me pull that or the caching changes out into their own PRs, I'm happy to do so.


src/main/grammars/ElmParser.bnf
```diff
@@ -392,7 +392,7 @@ Field ::=
 
 
 IfElse ::=
-    IF Expression THEN Expression ELSE Expression
+    IF Expression THEN Expression (ELSE IF Expression THEN Expression)* ELSE Expression
```
ajalt: This change matches how the Elm compiler parses these expressions, and makes them easier to work with.

src/main/grammars/ElmParser.bnf
```diff
@@ -472,7 +472,18 @@ ListPattern ::=
     LEFT_SQUARE_BRACKET [<<non_empty_list Pattern>>] RIGHT_SQUARE_BRACKET
 
 UnionPattern ::=
-    UpperCaseQID Pattern*
+    UpperCaseQID UnionArgumentPattern*
+
+private UnionArgumentPattern ::=
```
ajalt: Unparenthesized union patterns and `as` clauses can't appear as direct children of a union pattern.

src/main/kotlin/org/elm/ide/intentions/ElmMakeDeclarationIntentionAction.kt
```diff
@@ -47,11 +47,11 @@ class ElmMakeDeclarationIntentionAction : ElmAtCaretIntentionActionBase<ElmMakeD
         PsiDocumentManager.getInstance(project).doPostponedOperationsAndUnblockDocument(editor.document)
         editor.caretModel.moveCaretRelatively(0, 1, false, false, false)
 
-        val template = generateTemplate(project, editor, context)
+        val template = generateTemplate(project, context)
         TemplateManager.getInstance(project).startTemplate(editor, template)
     }
 
-    private fun generateTemplate(project: Project, editor: Editor, context: Context): Template {
```
ajalt: These parameters were unused

src/main/kotlin/org/elm/lang/core/psi/ElementTags.kt
```diff
@@ -27,17 +33,14 @@ interface ElmFunctionParamTag : ElmFunctionParamOrPatternChildTag
 /** An element that can be the direct child of an [ElmPattern] */
 interface ElmPatternChildTag : ElmFunctionParamOrPatternChildTag
 
-/** An element that can be the direct child of an [ElmConsPattern] */
-interface ElmConsPatternChildTag : ElmPsiElement
-
-/** An element that is either an [ElmTypeRefParameterTag], a [ElmParametricTypeRefParameterTag], or both. No elements implement this directly. */
-interface ElmTypeRefOrParametricTypeRefParameterTag : ElmPsiElement
```
ajalt: This tag ended up not being useful

src/main/kotlin/org/elm/lang/core/psi/PsiElement.kt
```diff
@@ -65,13 +63,6 @@ fun <T : PsiElement> PsiElement.parentOfType(vararg classes: KClass<out T>): T?
     return PsiTreeUtil.getParentOfType(this, *classes.map { it.java }.toTypedArray())
 }
 
-inline fun <reified T : PsiElement> PsiElement.parentsOfType(): Sequence<T> = parentsOfType(T::class.java)
-
-fun <T : PsiElement> PsiElement.parentsOfType(clazz: Class<out T>): Sequence<T> = parents().filterIsInstance(clazz)
-
-fun PsiElement.parents(): Sequence<PsiElement> = generateSequence(this) { it.parent }
```
ajalt: These functions were unused. `parents` is identical to `ancestors`. I can put `parentsOfType` back in if you think it will prevent someone from accidentally using the IntelliJ version that doesn't exist in WebStorm.

src/main/kotlin/org/elm/lang/core/psi/PsiElement.kt
```diff
@@ -81,21 +72,6 @@ inline fun <reified T : PsiElement> PsiElement.childOfType(strict: Boolean = tru
 inline fun <reified T : PsiElement> PsiElement.descendantsOfType(): Collection<T> =
         PsiTreeUtil.findChildrenOfType(this, T::class.java)
 
-fun PsiFile.descendantOfType(elementType: IElementType): PsiElement? {
```
ajalt: Removed in favor of `descendants`

ajalt:
The test failure looks like it might be a CircleCI problem? `gradlew check` runs fine locally.

klazuka:
> The test failure looks like it might be a CircleCI problem? gradlew check runs fine locally.

I think the problem is that it's running out of memory on the CircleCI box. Gradle is dying with exit code  137. See https://circleci.com/docs/2.0/java-oom/ and the recommendation at the end of https://stackoverflow.com/a/51978357

klazuka:
I have `JVM_OPTS: -Xmx3200m` in the circleci config yaml, which I think I got initially as part of their starter config for Java projects (or maybe I did it). In any case, based on that link above, I don't think it does what I expected it to do. So we should probably configure `GRADLE_OPTS` instead.

src/main/kotlin/org/elm/lang/core/resolve/scope/ModuleScope.kt
```diff
@@ -47,27 +63,32 @@ class ModuleScope(val elmFile: ElmFile) {
...
     fun getDeclaredValues(): List<ElmNamedElement> {
-        val valueDecls = elmFile.getValueDeclarations().flatMap {
-            it.declaredNames(includeParameters = false)
+        return CachedValuesManager.getCachedValue(elmFile, DECLARED_VALUES_KEY) {
+            val valueDecls = elmFile.getValueDeclarations().flatMap {
+                it.declaredNames(includeParameters = false)
+            }
+            val values = listOf(valueDecls, elmFile.getPortAnnotations(), elmFile.getInfixDeclarations())
+                    .flatten()
+            Result.create(values, elmFile.project.modificationTracker)
```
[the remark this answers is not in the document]
ajalt: If you change a declaration, then anything that references that declaration needs to be invalidated to avoid e.g. incorrect inference results. Since elements outside of this file can reference it, we need to invalidate everything.

src/main/kotlin/org/elm/lang/core/psi/ElmPsiManager.kt
```diff
@@ -0,0 +1,70 @@
...
+        private fun updateModificationCount(element: PsiElement) {
+            // If something is changed inside a function, we will only
+            // increment the function local modification counter. Otherwise, we will increment the
+            // global modification counter.
+
+            val owner = element.parentOfType<ElmValueDeclaration>()
+            if (owner == null) {
+                modificationTracker.incModificationCount()
+            } else {
+                owner.modificationTracker.incModificationCount()
```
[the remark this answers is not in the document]
ajalt: I believe this is safe because the PSI is mutable, and changes result in children being added and removed without affecting the parent element. The cache doesn't hold stale data, even logically. It holds references to elements that still exist in the psi, and have had some of their child elements changed.

src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
```diff
@@ -0,0 +1,923 @@
...
+        }
+        return inferReferencedValueDeclaration(ref?.parentOfType())
+    }
+
+    private fun inferReferencedValueDeclaration(decl: ElmValueDeclaration?): Ty {
+        if (decl == null || checkRecursion(decl)) return TyUnknown
+        val existing = resolvedDeclarations[decl]
+        if (existing != null) return existing
+        // Use the type annotation if there is one
+        var ty = decl.typeAnnotation?.typeRef?.let { typeRefType(it) }
```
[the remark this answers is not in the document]
klazuka: yes, and that would also be nice for IntelliJ's 'Extend/Shrink Selection' (option-up/down-arrow), which is based on the Psi tree structure. I frequently want to delete an entire function using Extend Selection, but currently it only grabs the `ElmValueDeclaration` part, leaving the type annotation dangling.

But this change can, and should, wait.

PR #109, opened 2018-11-01
Smarter insertion of virtual end decl tokens

klazuka:
Previously we were inserting them immediately after the last
code token, but if there was trailing whitespace or trailing
comment on that expression, that whitespace/comment would end
up AFTER the virtual end decl. Which is not desirable because
it logically belongs to the decl.

The Psi tree after this change is still not great in these
cases: the trailing whitespace is a _sibling_ of the value declaration.
But at least now it comes _before_ the virtual end decl.


src/main/kotlin/org/elm/lang/core/lexer/ElmLayoutLexer.kt
```diff
@@ -144,13 +144,23 @@ private fun doLayout(lexer: Lexer): ArrayList<Token> {
...
+                    //
+                    //   (1) trailing spaces at the end of the declaration are part of the declaration
+                    //   (2) top-level comments that follow the declaration are NOT part of the declaration
+                    //
+                    // Note that a virtual token has to appear after a whitespace token, since the real token
+                    // is combined with the virtual token during parsing (their text ranges overlap).
+                    loop@ for (k in (i - 1) downTo 1) {
+                        if (tokens[k].isCode) {
+                            for (m in (k + 1) until (i + 1)) {
+                                if (tokens[m].elementType == NEWLINE) {
```
ajalt: Will this cause problems if a file doesn't end in a newline?

klazuka: No, because virtual tokens are not emitted at EOF. At one time I had wanted to do that because it would make the parser (and recovery) cleaner, but I gave up on the dream.

PR #128, opened 2018-11-17
Propagate types information through parametric type refs

ajalt:
Currently, all type expressions are inferred as their declared types. For example, `List Int` is inferred as `List a`. This PR performs recursive substitution of argument types so that we can always infer the correct type of type expressions. This allows us to remove the workaround introduced in #124. Note that this does not change the behavior of value expressions. 

capture

I also added a new inspection that uses the propagated types to issue diagnostics. Currently, there is only one diagnostic, which checks argument count.

capture2

There are two other diagnostics that we could potentially add:

##### Enforcing type classes:

```elm
type A number = A
type alias B = A String
```

The Elm compiler does not issue diagnostics for this, so I chose not to as well.

##### Passing a non-record type into a record extension:

```elm
type alias A a = {a | x : ()}

foo : A Int -> ()
foo a = a.x
```

This actually [crashes the Elm compiler](https://github.com/elm/compiler/issues/1785). This is easy to check for, but in order to create a diagnostic for it, we would need to keep track of the psi element that the type came from.


klazuka:
FYI The Travis Windows build is broken because they updated their JDK version. I will fix it this weekend.

klazuka:
nice

PR #129, opened 2018-11-17
Setup an ElmProject when running tests; plug the "null ElmProject" hole

### LINKED ISSUE #119: Can't find references to elm-explorations/test when source-directories is ["."]
All the imports from `elm-explorations/test` are marked as `unresolved reference ''`.

Might it be because of the `elm-explorations/test` is under `test-dependencies` in `elm.json`?

`External Libraries` does have it:

image

Plugin version `2.0.1`

Anyhow - thank you so much for the plugin and taking time to develop it further!

klazuka:
Fixes #119 

In typical usage, every Elm file should have an associated Elm project (`elm.json`). But one big exception was our integration tests, which did not setup any Elm projects in the ElmWorkspace. And because of this hole, I had a fallback that would provide extremely liberal module resolution when the Elm project could not be determined. This fallback masked a lot of problems (including the underlying bug in #119).

This PR makes it so that all of the integration tests run in a workspace where an `ElmProject` exists. This makes the integration test code behave similarly to the real production code. And it allowed me to remove the loosey-goosey fallback in the module resolution code.

I also fixed the underlying bug in #119, which was that `test-dependencies` were not being included during module resolution.

Even after this PR, the module resolution system is still too liberal, but that will have to wait for a future PR.

### DIFF
--- a/src/test/kotlin/org/elm/ide/inspections/TypeInferenceInspectionTest.kt
+++ b/src/test/kotlin/org/elm/ide/inspections/TypeInferenceInspectionTest.kt
@@ -370,7 +370,7 @@ main : Maybe People
 main = person George
 --^
 
---@ Foo.elm
+--@ People/Washington.elm
 module People.Washington exposing (People(..), person)
 
 type People = George
--- a/src/main/kotlin/org/elm/lang/core/stubs/index/ElmModules.kt
+++ b/src/main/kotlin/org/elm/lang/core/stubs/index/ElmModules.kt
@@ -1,10 +1,9 @@
 package org.elm.lang.core.stubs.index
 
 import com.intellij.openapi.project.Project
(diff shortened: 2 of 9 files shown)


src/test/kotlin/org/elm/ide/inspections/TypeInferenceInspectionTest.kt
```diff
@@ -370,7 +370,7 @@ main : Maybe People
 main = person George
 --^
 
---@ Foo.elm
+--@ People/Washington.elm
```
klazuka: This was a bug in the original test, but it was being masked by my loosey-goosey fallback when the ElmProject was unknown (which it was in basically all of the tests, including this one).

klazuka:
@ajalt I built & tested this branch on my Windows 10 VM: all tests passed. So hopefully it won't break you.

PR #135, opened 2018-11-20
WIP Basic code completion for record fields

klazuka:
record field completion

Next up is to do completion when updating the field(s) in a record. Once that's in place, then we should have a good idea of what changes need to be made to the grammar.


src/main/kotlin/org/elm/lang/core/completion/ElmRecordFieldSuggestor.kt
```diff
@@ -0,0 +1,53 @@
...
+            //       `foo.bar.quux` => NOT supported
+            // TODO provide field completion when accessing a chain of fields
+            if (parent.directChildren.count { it.elementType == ElmTypes.DOT } > 1) return
+
+            // HACK: the type inference system does not yet keep track of the type
+            //       of the sub-expression before the dot character. So we will instead
+            //       just search for something in-scope with the same name.
+            // TODO figure out what changes need to made to the grammar / type system to make this better
+            val recordIdentifier = parent.lowerCaseIdentifierList.firstOrNull() ?: return
+            val recordTy = inference.expressionTypes.filter {
```
klazuka: I suspect that `expressionTypes` may contain things which are outside of the current lexical scope. So this is probably not the right thing to do.

ajalt: You are correct. The map includes all expressions and parameters in the top level declaration, so it can contain multiple values with the same name.

ajalt:
I think the best choice would be to combine the `FieldAccess` and `ExpressionAccessor` rules into something like `FieldAccess ::= DotWithoutWhitespace LOWER_CASE_IDENTIFIER`, and have each part of a chained access be its own expression/operand.

klazuka:
I'll try making that change to the grammar.

klazuka:
Abandoning in favor of separate PRs.

PR #138, opened 2018-11-23
Cleanup expression Psi

klazuka:
As part of trying to figure out how to do the left-recursive parse of `FieldAccess`, I found that there was some groundwork that needed to be put in place first. This PR does that by changing the way that the `Operand` rule is defined so that it triggers GrammarKit's special expression parsing mode which can handle left-recursive rules (with restrictions). ([more details](https://github.com/JetBrains/Grammar-Kit/issues/95#issuecomment-189475877))

One immediate benefit is that GrammarKit will present nicer errors in some cases.

Before:
before

After:
after

Notable changes:
- parenthesized expressions now have their own Psi element in the tree
- `Operand` is now called `Atom`. It seemed strange to call something an operand in the case where it's not even being applied to a function/operator.
- all "atoms" now have an `Expr` suffix, this was needed to make it easier to write/maintain GrammarKit's `extends` statement (necessary for the expression/Pratt parser mode)


klazuka:
I'm pretty sure that this is the way to go (and it does roughly correspond to how intellij-rust handles such things), but I'm a little reluctant about relying on additional GrammarKit functionality.

klazuka:
One thing I didn't try is moving the `extends` all the way up to our `Expression` rule (which would probably be more correct). I'm not sure if it will satisfy GrammarKit's restrictions, but I will give it a try.

klazuka:
@ajalt FYI: I just pushed an experimental commit that takes GrammarKit's special expression parsing mode all the way up to our main `Expression` rule. There's some breakage that I haven't had the time to fix yet—and I still need to reconcile some things like the various expression tag interfaces—but I wanted to put this up on Github since it's a pretty big change. It makes the Psi *much* cleaner, getting rid of noise elements like `ElmExpression`.

And another big bonus is that the expression parsing rule is much more efficient now. Given an expression like `List.length [a,b,c]`, previously it would parse the function call plus its operand and then try to look for a binary operator token. That would fail and it would then backtrack and try the alternative which re-parses the function call. Now I'm using an `upper` rule which lets you parse that common prefix once, and then depending on what tokens follow, change the element type dynamically.

src/main/kotlin/org/elm/lang/core/psi/elements/ElmBinOpExpr.kt
```diff
@@ -0,0 +1,16 @@
...
+import org.elm.lang.core.psi.ElmExpressionTag
+import org.elm.lang.core.psi.ElmPsiElementImpl
+import org.elm.lang.core.psi.directChildren
+
+
+/**
+ * One or more binary operator expressions
+ */
+class ElmBinOpExpr(node: ASTNode) : ElmPsiElementImpl(node), ElmExpressionTag {
+    val operators: Sequence<ElmOperator> get() = directChildren.filterIsInstance<ElmOperator>()
```
ajalt: You'll need to replace these two properties with a single sequence of both operators and atoms:
```kotlin
/** The operands and operators in this expression */
val parts: Sequence<ElmExpressionPartTag> get() = directChildren.filterIsInstance<ElmExpressionPartTag>()
```

Iterating over them separately isn't usually useful.

klazuka:
@ajalt FYI notable changes in recent commits:
- I introduced a couple more tags
    - `ElmBinOpPartTag` which includes binary operators and their operands
    - `ElmOperandTag` either a function call expression or an "atom" expr
- prior to this PR, function call expressions had the old "Operand" tag (now "Atom") even though they were not parsed as part of that rule. I thought that was misleading, so now it's a separate thing. This required a small change in type inference so that inferring a function call expression stores the result in the inference result's `expressionTypes`.
- I added a bare-bones test for `ElmExpressionTypeProvider`
- I now exclude parenthesized expressions from the list of choices given by the expression type provider. The other option was to simply not register `ElmParenthesizedExpr` in the inference result's `expressionTypes`, but it seemed more correct to do it in the expr type provider.

klazuka:
Assuming this is ok with you, I'm going to merge it. And then I will start working on the left-associative parsing of field access expressions.

PR #140, opened 2018-11-26
Parse FieldAccessExpr as left-associative

klazuka:
Code completion of record field names in expressions like `person.name.first` requires being able to infer the type of each sub-expression in the chain. Previously we were parsing it as a single node and thus we were only able to cache a single inferred type for the expression.

Now we do a left-associative parse of field access expressions and type inference now registers the type of each sub-expression.

I will do the actual code completion work in a subsequent PR.


src/main/grammars/ElmParser.bnf
```diff
@@ -138,11 +139,9 @@ external DotWithoutTrailingWhitespace ::= parseDotWithoutTrailingWhitespace
 external DotWithoutLeadingWhitespace ::= parseDotWithoutLeadingWhitespace
 external MinusWithoutTrailingWhitespace ::= parseMinusWithoutTrailingWhitespace
 
-UpperCaseQID ::= UPPER_CASE_IDENTIFIER (DotWithoutWhitespace UPPER_CASE_IDENTIFIER)* !DotWithoutLeadingWhitespace
```
klazuka: I removed the `!DotWithoutLeadingWhitespace` lookahead because it no longer seems to be necessary (probably due to the grammar changes that I made earlier today).

src/main/kotlin/org/elm/lang/core/psi/elements/ElmFieldAccessExpr.kt
```diff
@@ -2,49 +2,22 @@ package org.elm.lang.core.psi.elements
...
+        get() = findNotNullChildByClass(ElmFieldAccessPartTag::class.java)
 
-    /* TODO [kl] a more correct implementation would filter the visible values
-       to those that we know are actually records. But we haven't implemented
-       any kind of type system in the plugin yet. */
-    override fun getVariants(): Array<ElmNamedElement> =
-            ExpressionScope(element).getVisibleValues()
-                    .toTypedArray()
-}
+    val lowerCaseIdentifier: PsiElement
```
klazuka: This should eventually become a PsiReference to the field in the corresponding record type declaration, but I'm hesistant to do that work right now.

src/main/grammars/ElmParser.bnf
```diff
@@ -314,11 +311,9 @@ private Atom ::=
 // Various tokens that can begin a new expression. Useful for parse error recovery.
 private expr_delimiters ::= COMMA | LEFT_BRACE | LEFT_PARENTHESIS | LEFT_SQUARE_BRACKET
 
-ExpressionWithAccessorExpr ::=
-    ParenthesizedExpr ExpressionAccessor
-
-RecordWithAccessorExpr ::=
-    RecordExpr ExpressionAccessor
+FieldAccessExpr ::= FieldAccessStart FieldAccessSegment+
+FieldAccessStart ::= LOWER_CASE_IDENTIFIER | ParenthesizedExpr | RecordExpr
```
ajalt: I don't know if this is viable, but the Psi would probably be easier to work with if, rather than creating this wrapper rule, we did something like this:
```
private FieldAccessStart ::= ValueExpr | ParenthesizedExpr | RecordExpr
```

src/main/kotlin/org/elm/lang/core/psi/elements/ElmFieldAccessExpr.kt
```diff
@@ -2,49 +2,22 @@ package org.elm.lang.core.psi.elements
...
-
-
-class BaseRecordReference(element: ElmReferenceElement)
-    : ElmReferenceCached<ElmReferenceElement>(element) {
+class ElmFieldAccessExpr(node: ASTNode) : ElmPsiElementImpl(node), ElmAtomTag, ElmFunctionCallTargetTag, ElmFieldAccessPartTag {
 
-    override fun resolveInner(): ElmNamedElement? {
-        return getVariants().find { it.name == element.referenceName }
-    }
+    val leftExpr: ElmFieldAccessPartTag
```
ajalt: If I understand this correctly, this is `model.currentUser` in `model.currentUser.name`? Maybe naming it `target` or `targetExpr` to match `ElmFunctionCall` would be more clear.

src/main/kotlin/org/elm/lang/core/psi/elements/ElmFieldAccessStart.kt
```diff
@@ -0,0 +1,63 @@
...
+        get() = findChildByType(ElmTypes.LOWER_CASE_IDENTIFIER)
+
+    val recordExpr: ElmRecordExpr?
+        get() = findChildByClass(ElmRecordExpr::class.java)
+
+    val parenExpr: ElmParenthesizedExpr?
+        get() = findChildByClass(ElmParenthesizedExpr::class.java)
+
+    /*
+        NOTE: due to how we've defined ElmReferenceElement, a Psi element
```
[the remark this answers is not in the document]
ajalt: In the bnf file, I suggested that this rule be private. If that's possible, then this problem would go away.

src/main/kotlin/org/elm/lang/core/psi/elements/ElmFieldAccessExpr.kt
klazuka: 👍

PR #141, opened 2018-11-27
Add code completion for accessing fields in a record

klazuka:
Now you can do code completion after typing `.` following a record value:
screen shot 2018-11-26 at 4 35 35 pm

This PR does __not__ provide completions for the record literal update syntax (e.g. `{ user | name = "" }`). Implementing that is a bit tricky, so it will have to wait until later.


src/test/kotlin/org/elm/lang/core/completion/ElmCompletionTest.kt
```diff
@@ -160,7 +155,14 @@ g = Data{-caret-}
 """)
 
 
-    fun `test value completion of module prefix, after dot`() = doSingleCompletionMultiFile(
+/*
+TODO re-enable these tests once we figure out how to intelligently pick
```
klazuka: I intentionally punted on this for now. The tradeoff is worth it.

PR #151, opened 2018-11-30
Various performance fixes (and some bug fixes and cleanup along the way)

klazuka:
Profiling with YourKit showed a lot of time being spent in the various `ModuleScope` methods as well as `ElmFile.getImportDecls()`. I fixed the former by adding more caching, and the latter by changing how it walks the Psi tree.

Along the way, I fixed a bug related to module resolution and refactored the code so that it couldn't happen again. And I cleaned up some other hacks that are no longer necessary now that we model Elm projects.


klazuka:
I've been running this all day—performance is much better and memory usage doesn't look much higher.

PR #155, opened 2018-12-01
 Aliased imports now hide the original module name from qualified refs

### LINKED ISSUE #93: Importing a module with an alias should 'hide' the full module name during name resolution
Thanks for creating this plugin.

I just tried the plugin and noticed that qualified imports are not supported. For example:
```
import Html as H
```

it looks like plugin only "sees" the `import Html`

klazuka:
Fixes #93 

When a module is imported using an alias (e.g. `import Json.Decode as D`), Elm prohibits the use of the original module name in qualified references. We now mark these as unresolved references along with a special error message that describes the problem and its solution.

### DIFF
--- a/src/main/kotlin/org/elm/ide/annotator/ElmUnresolvedReferenceAnnotator.kt
+++ b/src/main/kotlin/org/elm/ide/annotator/ElmUnresolvedReferenceAnnotator.kt
@@ -6,24 +6,52 @@ import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiReference
 import org.elm.ide.intentions.ElmImportIntentionAction
 import org.elm.ide.intentions.ElmMakeDeclarationIntentionAction
+import org.elm.lang.core.psi.ElmFile
+import org.elm.lang.core.psi.ElmQID
 import org.elm.lang.core.psi.ancestors
 import org.elm.lang.core.psi.elements.*
+import org.elm.lang.core.resolve.reference.QualifiedConstructorReference
+import org.elm.lang.core.resolve.reference.QualifiedTypeReference
+import org.elm.lang.core.resolve.reference.QualifiedValueReference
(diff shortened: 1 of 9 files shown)


src/main/kotlin/org/elm/ide/annotator/ElmUnresolvedReferenceAnnotator.kt
```diff
@@ -6,32 +6,69 @@ import com.intellij.psi.PsiElement
...
+            ::handleTypeAnnotation,
+            ::handleSafeToIgnore,
+            ::handleModuleHiddenByAlias
+    )
+
     override fun annotate(element: PsiElement, holder: AnnotationHolder) {
         for (ref in element.references) {
             if (ref.resolve() == null) {
+                var handled = false
+                for (handler in handlers) {
```
ajalt: This could potentially be shortened to 
```kotlin
if (ref.resolve() == null && handlers.none { it(ref, element, holder) }) {
    holder.createErrorAnnotation(...)
}
```

src/main/kotlin/org/elm/ide/annotator/ElmUnresolvedReferenceAnnotator.kt
```diff
@@ -6,32 +6,69 @@ import com.intellij.psi.PsiElement
...
+ */
+typealias BadRefHandler =
+        (ref: PsiReference, element: PsiElement, holder: AnnotationHolder) -> Boolean
+
 
 class ElmUnresolvedReferenceAnnotator : Annotator {
 
+    // A chain of handlers to be executed sequentially when a reference cannot be resolved.
+    // Order matters! Handlers earlier in the chain can short-circuit further evaluation.
+    private val handlers = listOf<BadRefHandler>(
```
ajalt: I like this approach: simple and scalable :+1:

PR #156, opened 2018-12-02
Do code completion for functions & types even when module not imported

### LINKED ISSUE #29: Do code completion for modules that are not imported
Let's say that I have not yet imported `List.Extra`. Now I start typing `List.Extra.` in an expression. We should provide code completions from that module.

As a bonus, it would be nice to automatically import the module when you select the completion.

klazuka:
Fixes #29 

Things left to do before merging:

- [x] fix auto-import intention for ParametricTypeRef: generates the wrong import
- [ ] try to make code-completion after the dot for upper-case ids work again


klazuka:
I punted on trying to re-enable code-completion after the dot for upper-case ids. This broke when I did the record field completion work, and it looks like fixing it will require some parse error recovery work so that we have a nicer Psi tree in `ElmCompletionContributor.beforeCompletion`.

PR #163, opened 2018-12-14
Improve type expression inference algorithm

ajalt:
This PR contains a number of changes related to type expression inference. The largest is a new approach to handling arguments in parameterized type expressions. An explanation of this algorithm was added to `TypeExpression`; let me know if it's still unclear or could be improved.

As part of this work, I made some improvements to the `Ty` classes:

* I added information about aliases to all types. It was previously only available for records. 
* I added information about union members to `TyUnion`.

Neither of those changes affect value inference, but they add enough information to the `Ty`s that the quick docs can rely entirely on `TypeRendering` rather than using separate rendering code for type references. This allows us to remove a significant chunk of the documentation provider code, and finally get rid of the last rendering differences between quick docs and the expression type provider. This also allowed me to add quick docs for union members/constructors.

The original impetus for the algorithm changes was to be able to cache the inference result for type declarations and aliases. However, this PR doesn't actually add any caching. For reasons I don't understand, when I tried to use the `CachedValuesManager` in the same way I did for value inference, IntelliJ would throw exceptions related to invalid PSI references. We may want to revisit this caching in the future.


klazuka:
My dad is visiting this weekend. I will try to look at this PR in the next couple of days.

PR #165, opened 2018-12-16
Fix list pattern binding

### LINKED ISSUE #164: List destructuring shows error for 2 items
This is the problem 
image

The type checker thinks that `height` is the list's tail instead of the second item in a list of 2 items. 
I can compile without problem and the program executes as expected.

When I try to alter the destructuring into

image

It doesn't display an error anymore but the program behaves not as expected.

The parseWithDefault function 
image

What can I do?
Thank you!

ajalt:
Fixes #164


klazuka:
Looks good.

Travis Linux builds are failing because they got rid of the container-based "sudo" system, and, apparently, npm is not provided in the new environment. I'll look into it later this week.

PR #168, opened 2018-12-17
Add inspection and enter handler to complete case branches

### LINKED ISSUE #160: Provide an intention action to populate the branches in a `case` expression
One little gadget from elmjutsu in Atom that I like best is that if you type `case var of` in a function it can resolve the type and the various required cases and generates a skeleton of all the cases.

I'm suggesting (begging for?) it to be added here.  Say I have:

```elm
type Msg
    = Foo String
    | Bar 
```

Then I setup:

```elm
update : Msg -> Model -> ( Model, Cmd Msg)
update msg model = 
    case msg of
```

Once I complete `of` I automagically get
```elm
update : Msg -> Model -> ( Model, Cmd Msg)
update msg model = 
    case msg of
        Foo str -> -- `str` is part of the autogeneration

        Bar ->
```
Makes me smile every time.  

Thanks for the great tool and thanks for reading this far.

ajalt:
Add inspection and enter handler to complete case branches

This PR uses the new member property on `TyUnion` to show an inspection when a case expression on a union type is missing one or more variant patterns. 

There is a new inspection that will annotate the `case` keyword and offer quick fixes to add the missing branches, or add a wildcard branch.

untitled

There is also a new SmartEnterProcessor that implements the "complete current statement" action (bound by default to Ctrl+Shift+Enter) to add missing branches without needing to select a quick fix. We could add other expressions like `if` and `let` to the processor in the future.

In order to implement this, I made a minor change to the grammar: I made the `CaseOfTail2` rule optional so that no error element is generated if you haven't added any branches yet. This was necessary to get the type of the case expression, since we don't infer functions that have errors. The "missing case branches" error is more user friendly than the parse error anyway, although we don't yet have exhaustive pattern checking.

I chose not to add any value to the generated branches. We could add `Debug.crash "TODO"`, but that would require adding an import, and I didn't want to break anyone's build when then inevitably forget to remove that import. Once we can detect unused imports, it might be worth adding.

Fixes #160


src/main/kotlin/org/elm/lang/core/types/TypeRendering.kt
```diff
@@ -55,3 +55,25 @@ private fun StringBuilder.renderLink(refText: String, text: String, linkify: Boo
...
+        is TyTuple -> renderParam()
+        is TyVar -> name
+        is TyUnit -> "()"
+        is TyUnknown, TyInProgressBinding -> "unknown"
+        TyShader -> "shader"
+    }
+}
+
+fun TyTuple.renderParam(): String {
+    return types.joinToString(", ", prefix = "(", postfix = ")") { it.renderParam() }
```
ajalt: I chose to destructure tuples in params since that seems pretty common, but I'm happy to just render `tuple` if you think we should be consistent with the other types.

src/test/kotlin/org/elm/ide/intentions/ElmMakeDeclarationIntentionTest.kt
```diff
@@ -34,21 +35,25 @@ f : (Int -> Int) -> List a -> (Char, String) -> { foo : Int } -> Bool{-caret-}
...
+f function list (char, string) record = {-caret-}
 """)
     }
 
 
     fun `test function parameters should be camelCased`() {
         doAvailableTest(
                 """
+type FooBar = FooBar
+type QuuxQuuxQuux = QuuxQuuxQuux
```
ajalt: The downside to using the `Ty` to generate params is that we don't have any information if the type isn't resolved. This also pops up occasionally in documentation rendering. Maybe I could add a new property to `TyUnknown` that contains the name of unresolved references so that we can show them anyway.

klazuka:
FYI, Elm imports the `Debug` module implicitly:
screen shot 2018-12-17 at 11 09 40 am

klazuka:
> In order to implement this, I made a minor change to the grammar: I made the CaseOfTail2 rule optional so that no error element is generated if you haven't added any branches yet. This was necessary to get the type of the case expression, since we don't infer functions that have errors. The "missing case branches" error is more user friendly than the parse error anyway, although we don't yet have exhaustive pattern checking.

Cool. This whole `CaseOfTail2` thing is a mess of my own creation. It all resulted from me flailing around, doing whatever it took to get parse error recovery working reasonably well. Perhaps now that we have relaxed things a bit, maybe the grammar can be simplified.

src/main/kotlin/org/elm/ide/inspections/MissingCaseBranchAdder.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    fun addWildcardBranch() {
+        if (missingBranches.isEmpty()) return
+        addPatterns(listOf("_"))
+    }
+
+    private fun addPatterns(patterns: List<String>) {
+        val factory = ElmPsiFactory(element.project)
+        val existingBranches = element.branches
+        val indent = "    ".repeat(element.ancestors.takeWhile { it !is ElmFile }.count())
```
klazuka: Clever. There are probably cases where this will break, but this is a great solution in lieu of doing the full code-formatter integration.

PR #182, opened 2018-12-27
Intentions for adding/removing a function/type from the exposing list

### LINKED ISSUE #147: Add/remove a value/function/type from the module's exposing list
There are 2 steps to this:
1. add an intention action on top-level things to add it to the module's exposing list
2. add an action to the "exposed" gutter icon to remove it from the exposing list

One complication is that messing around with the exposing list is kind of a pain because you have to manually deal with the PsiWhitespace elements. It'd be nice if we had a better way to do local code formatting for whitespace.

As for (2), I found the following tip on the IntelliJ dev forums:
> The renderer returned by LineMarkerInfo instances you generate should define popup menu actions (com.intellij.openapi.editor.markup.GutterIconRenderer#getPopupMenuActions). You can look at RunLineMarkerProvider, as an example (that's what shows popup in your screenshot).

klazuka:
Fixes #147 

New stuff:
1. "add to exposing list" intention
2. "remove from exposing list" intention
3. the "e" gutter line marker icon is now clickable: it takes you to where the thing is exposed in the exposing list

add

remove

I also consolidated the logic for determining if a value/function/type is exposed and added some utility functions for creating/deleting stuff in an `exposing` list. These utility functions will be used in a subsequent pull request to cleanup the "add import" intention code.

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -95,22 +95,26 @@
         <additionalTextAttributes scheme="Darcula" file="colorSchemes/ElmDarcula.xml"/>
         <annotator language="Elm" implementationClass="org.elm.ide.highlight.ElmSyntaxHighlightAnnotator"/>
         <annotator language="Elm" implementationClass="org.elm.ide.annotator.ElmUnresolvedReferenceAnnotator"/>
-        <colorSettingsPage implementation="org.elm.ide.color.ElmColorSettingsPage" />
+        <colorSettingsPage implementation="org.elm.ide.color.ElmColorSettingsPage"/>
         <enterHandlerDelegate implementation="org.elm.ide.typing.ElmOnEnterSmartIndentHandler"/>
         <fileTypeFactory implementation="org.elm.lang.core.ElmFileTypeFactory"/>
-        <gotoSymbolContributor implementation="org.elm.ide.navigation.ElmGoToSymbolContributor" />
-        <lang.braceMatcher language="Elm" implementationClass="org.elm.ide.ElmPairedBraceMatcher" />
+        <gotoSymbolContributor implementation="org.elm.ide.navigation.ElmGoToSymbolContributor"/>
+        <lang.braceMatcher language="Elm" implementationClass="org.elm.ide.ElmPairedBraceMatcher"/>
         <lang.commenter language="Elm" implementationClass="org.elm.ide.commenter.ElmCommenter"/>
-        <completion.contributor language="Elm" implementationClass="org.elm.lang.core.completion.ElmCompletionContributor" />
-        <lang.findUsagesProvider language="Elm" implementationClass="org.elm.ide.search.ElmFindUsagesProvider" />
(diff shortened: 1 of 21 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -95,30 +95,35 @@
         <additionalTextAttributes scheme="Darcula" file="colorSchemes/ElmDarcula.xml"/>
         <annotator language="Elm" implementationClass="org.elm.ide.highlight.ElmSyntaxHighlightAnnotator"/>
         <annotator language="Elm" implementationClass="org.elm.ide.annotator.ElmUnresolvedReferenceAnnotator"/>
-        <colorSettingsPage implementation="org.elm.ide.color.ElmColorSettingsPage" />
+        <colorSettingsPage implementation="org.elm.ide.color.ElmColorSettingsPage"/>
```
klazuka: IntelliJ did a bunch of whitespace reformatting in this file. Ignore.

src/main/resources/META-INF/plugin.xml
```diff
@@ -139,22 +144,37 @@
                          implementationClass="org.elm.ide.inspections.ElmIncompletePatternInspection"/>
 
 
+        <!-- Intentions -->
+
+        <intentionAction>
```
klazuka: These are the real changes to the `plugin.xml`. When you register an intention action like this, you also need to provide documentation HTML and before/after examples. Failure to do so will result in IntelliJ disabling your plugin at launch time. So I added a new test `ElmIntentionDocsTest` (based on intellij-rust), which verifies that the documentation exists.

src/test/kotlin/org/elm/ide/intentions/ElmIntentionDocsTest.kt
```diff
@@ -0,0 +1,27 @@
+package org.elm.ide.intentions
+
+import com.intellij.codeInsight.intention.IntentionManager
+import org.elm.lang.ElmTestBase
+
+class ElmIntentionDocsTest : ElmTestBase() {
```
klazuka: All intentions registered in the `plugin.xml` must be documented.

src/test/kotlin/org/elm/ide/intentions/ElmIntentionTestBase.kt
```diff
@@ -8,13 +8,11 @@
...
 
 import com.intellij.codeInsight.intention.IntentionAction
-import org.elm.ide.inspections.ElmAnnotationTestBase
 import org.elm.lang.ElmTestBase
-import org.elm.replaceCaretMarker
 import org.intellij.lang.annotations.Language
 
 
-abstract class ElmIntentionTestBase(val intention: IntentionAction) : ElmAnnotationTestBase() {
+abstract class ElmIntentionTestBase(val intention: IntentionAction) : ElmTestBase() {
```
klazuka: This was an early mistake that I made. It should have never inherited from `ElmAnnotationTestBase`

PR #185, opened 2018-12-28
Allow exponent in number literal

alvinlindstam:
Elm allows `123e10`, `123.0e10` and `123e-10` and such patterns as number literals, but the plugin does not recognise the syntax. 

(I'm not sure if this is the correct fix, or if it's sufficient, but it seemed like the place to start)


klazuka:
Yep, you're right. This was an oversight. Would you mind adding a test to `ElmBasicLexerTest`? And then I'll merge it. Thanks!

PR #186, opened 2018-12-29
Add inspections for unused functions, parameters, imports, etc.

### LINKED ISSUE #148: Gray-out declarations that have no incoming references

klazuka:
Fixes #148 and #149 

In a subsequent PR I will add 'optimize imports' fix/intention.

I did some performance testing in PowerLift's massive `Main.elm` file because I was worried that detecting unused functions would be slow. But it's fast. 

Making it fast required relaxing the unused function check to exclude cases where there are a lot of candidates (e.g. a function named `view`), so there will be some false negatives.

I also expect that there will be some false positives, but initial testing using elm-spa-example and PowerLift both look ok. So I think it's ready to ship.


src/main/kotlin/org/elm/ide/inspections/ElmUnusedImportInspection.kt
```diff
@@ -0,0 +1,78 @@
...
+                "'${exposedItem.text}' is unused",
+                ProblemHighlightType.LIKE_UNUSED_SYMBOL
+        )
+    }
+}
+
+class ImportVisitor(initialImports: List<ElmImportClause>) : PsiElementVisitor() {
+
+    // IMPORTANT! IntelliJ's LocalInspectionTool requires that visitor implementations be thread-safe.
+    private val imports = initialImports.toMutableList()
```
klazuka: Now that I think about it, maybe I should use a ConcurrentHashMap indexed by the module name string.

src/main/kotlin/org/elm/ide/inspections/ElmUnusedImportInspection.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    private fun markAsUnused(holder: ProblemsHolder, importClause: ElmImportClause) {
+        holder.registerProblem(
+                importClause,
+                "Unused import",
+                ProblemHighlightType.LIKE_UNUSED_SYMBOL
+        )
+    }
+
+    private fun markAsUnused(holder: ProblemsHolder, exposedItem: ElmExposedItemTag) {
```
klazuka: whoops, I forgot to implement this. I'll do it later when I do optimize imports.

PR #190, opened 2019-01-01
 Improve name resolution for test modules and dependencies

### LINKED ISSUE #189: Module in "tests" directory cannot be imported by other files in "tests"
With elm 0.19.0 I've modularized my code like this:

```
src/ModuleA
src/ModuleB -- imports ModuleA
tests/ModuleATest -- imports ModuleA, ModuleASupport
tests/ModuleASupport
tests/ModuleBTest -- imports ModuleB, ModuleA, ModuleASupport
```

On the command-line, both `elm make` and `elm-test` run without problems.
In IntelliJ (2018.3.2 CE) with the 'Elm' plugin (v2.1.1) and the 'Elm External Tools' plugin (v2.0.1), all the references to functions in `tests/ModuleASupport` are marked as "Unresolved reference" though.

I'm not sure whether this is an issue with the 'Elm' plugin, or with the 'Elm External Tools' plugin.  
I decided to post this issue here, because the behavior doesn't change if I disable the 'Elm External Tools' plugin.

As my project structure is valid for both `elm make` and `elm-test`, I hope there is a way to also make the 'Elm' plugin understand it.

klazuka:
Notable changes:

- the "tests" directory is now treated as a source-root when the calling code is also in the "tests" directory (fixes #189)
- "test-dependencies" are now only included in code completion and reference resolve results when the calling code is in the "tests" directory. Previously they were always included, which was a bug.


src/main/kotlin/org/elm/lang/core/psi/ElmFile.kt
```diff
@@ -124,3 +136,20 @@ class ElmFile(viewProvider: FileViewProvider) : PsiFileBase(viewProvider, ElmLan
...
+ * Describes the location from which a reference is resolved.
+ *
+ * Reference resolution depends on context. For instance, we need
+ * to know the containing Elm project in order to determine which
+ * `source-directories` are valid roots.
+ *
+ * Starting in Elm 0.19, the Elm project's `test-dependencies`
+ * are only "import-able" from within "$ProjectRoot/tests" directory.
+ */
+interface ClientLocation {
```
klazuka: Not sure about this name. Maybe it would be better as `RefContext`?

I also wonder if it's a bit too clever making `ElmFile` implement this interface. Maybe it'd be better to make things more explicit by giving `ClientLocation` a factory function that takes an `ElmFile`.

PR #193, opened 2019-01-03
Add caching for type expression inference

### LINKED ISSUE #191: Performance degradation in long files
I noticed that the longer a file is, the longer IntelliJ's analysis phase takes. I also noticed that typing is a bit sluggish in large files. At first I thought it might have been caused by new inspections for finding unused code and unused imports, but disabling them did not fix the problem.

It appears that the problem is with the 'missing case branches' inspection or maybe one of the type checker inspections (or some combination thereof). Here are my raw notes on digging in to the problem:

All tests were performed using PowerLift's `src/App/Main.elm` file. All timings measured on my laptop on battery power w/ 30% juice. Each test measured the time from pressing a key until IntelliJ's analysis completed.

### Results
w/ everything enabled: ~13 seconds
w/ all optional inspections disabled (but still using the unresolved ref annotator): ~5 sec
w/ Unused Import enabled: < 2 seconds
w/ Unused Symbol enabled: < 2 seconds
w/ both 'Unused $foo' inspections' enabled: < 2 seconds

this doesn't make any sense other than that there is some kind of caching happening when we
use the unused* stuff that's not happening with the unresolved ref annotator.

w/ 'Missing Case Branches' enabled: **~14 seconds (BAD!!!!!!!!!)**
w/ 'Type Checker' (`ElmTypeInferenceInspection`) enabled: **48 seconds** (but I think this is exceptionally slow because it depends on inference results which aren't getting cached)
w/ 'Type Declaration Checker' (`ElmTypeDeclarationInspection`) enabled: 7 seconds (acceptable considering baseline)

-- 

cc @ajalt 
I might look into this this week. While I think the perf regression is already present in 2.1.1, I'm going to hold off on releasing 2.2.0 until we figure out the perf issue.

klazuka:
Fixes #191 

I haven't run into any exceptions with this yet. And it improved the analysis performance on PowerLift's `App/Main.elm` from ~13 seconds down to 5 seconds.

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/types/TypeExpressionInference.kt
+++ b/src/main/kotlin/org/elm/lang/core/types/TypeExpressionInference.kt
@@ -1,34 +1,57 @@
 package org.elm.lang.core.types
 
+import com.intellij.openapi.util.Key
+import com.intellij.psi.util.CachedValue
+import com.intellij.psi.util.CachedValueProvider
+import com.intellij.psi.util.CachedValuesManager
 import org.elm.lang.core.diagnostics.BadRecursionError
 import org.elm.lang.core.diagnostics.ElmDiagnostic
 import org.elm.lang.core.psi.ElmNameIdentifierOwner
 import org.elm.lang.core.psi.ElmTypeSignatureDeclarationTag
 import org.elm.lang.core.psi.elements.*
+import org.elm.lang.core.psi.modificationTracker
 import org.elm.lang.core.psi.parentOfType
 
 
-// Changes to type expressions aways invalidate the whole project, since they influence inferred
+// Changes to type expressions always invalidate the whole project, since they influence inferred
 // value types (e.g. removing a field from a record causes usages of that field everywhere to be invalid.)
-// These results should be able to be cached, but doing so with the CachedValueManager lead to invalid Psi references.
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/elm/lang/core/types/TypeExpressionInference.kt
```diff
@@ -1,34 +1,57 @@
...
+            val inferenceResult = TypeExpression().beginUnionConstructorInference(this)
+            CachedValueProvider.Result.create(inferenceResult, project.modificationTracker)
+        }
+
 
 /** Get the type of the expression in this annotation, or null if the program is incomplete and no expression exists */
 fun ElmTypeAnnotation.typeExpressionInference(): ParameterizedInferenceResult<Ty>? {
     val typeRef = typeExpression ?: return null
-    return TypeExpression().beginTypeRefInference(typeRef)
+    return CachedValuesManager.getCachedValue(typeRef, TY_CACHE_KEY) {
```
klazuka: At first I passed `this` as the first argument to `getCachedValue`, and I *did* see an exception from IntelliJ saying that I was doing it wrong. But once I changed it to `typeRef`, it appears to work correctly.

klazuka:
I just need to dig in and understand more deeply how IntelliJ's caching system works. But this seems good enough for now.

PR #211, opened 2019-01-10
Improve case branch completion

### LINKED ISSUE #192: Add missing case branches should use qualified refs when appropriate
When pattern-matching on a union type's variants using qualified refs, the 'add missing case branches' intention should also generate code that uses qualified refs.

For example:

File `Foo.elm`
```elm
module Foo exposing (Foo(..))
type Foo = Bar | Baz | Quux
```

File `Main.elm`
```elm
import Foo
f x = case x of
               Foo.Bar -> ()
               Foo.Baz -> ()
```

Running the 'add missing case branches' intention should generate a line like:
```elm
               Foo.Quux ->
```

But it actually generates
```elm
               Quux ->
```
which is in an unresolved reference because the union variants were not exposed by the import.

ajalt:
This PR has two commits:

1. Qualifies generated union variants when necessary (i.e. the module import doesn't expose the union type)
2. Shows the wildcard fix for all case expressions without branches. Previously no error was shown unless the expression was a union, even if the case has no branches.

Fixes #192


julianpistorius:
Just a quick FYI: I am pretty sure this PR introduced a bug where it thinks that branches are incomplete, but it's actually fine. I'm still trying to reproduce this.

PR #215, opened 2019-01-12
Problem: False positive for non-exhaustive case expression

### LINKED ISSUE #214: False positive for non-exhaustive case expression
It seems PR #211 introduced a bug where it determines a case expression is non-exhaustive when it is actually exhaustive.

I built and tested both the commit that merged the PR (510a3e08f36ef32da388a056342e447798b1588d) and the one before (5cc0a46d20e90dc3e9393b7322672d97318b0232)

I made a simple Elm file with a `case` with a `Bool`:

```elm
module CaseTest exposing (..)

simpleCase : Bool -> String
simpleCase aBoolThing =
    case aBoolThing of
        True ->
            "I think it's True"

        False ->
            "I think it's False"
```

Screenshot with plugin built from 510a3e08f36ef32da388a056342e447798b1588d showing false positive:

screenshot_510a3e0

Screenshots with plugin built from 5cc0a46d20e90dc3e9393b7322672d97318b0232 not showing the false positive, as well as showing a true positive:

screenshot_5cc0a46_02

screenshot_5cc0a46_01

julianpistorius:
PR #211 introduced a bug where it determines a case expression
is non-exhaustive when it is actually exhaustive.

Solution: If missing branches is empty then no branches are missing

(I feel really silly writing that :D)

Fixes #214

### DIFF
--- a/src/main/kotlin/org/elm/ide/inspections/MissingCaseBranchAdder.kt
+++ b/src/main/kotlin/org/elm/ide/inspections/MissingCaseBranchAdder.kt
@@ -96,6 +96,10 @@ class MissingCaseBranchAdder(val element: ElmCaseOfExpr) {
             }
         }
 
+        if (missingBranches.isEmpty()) {
+            return Result.NoMissing
+        }
+
         val qualifierPrefix = ModuleScope(element.elmFile).getQualifierForTypeName(exprTy.module, exprTy.name) ?: ""
 
         return Result.MissingVariants(missingBranches.mapKeys { (k, _) -> qualifierPrefix + k })


julianpistorius:
I should really make a test that fails in the presence of this bug that I (in theory) fixed.

klazuka:
@julianpistorius thanks for figuring this out and contributing the fix

viktor-ferenczi:
Figured what can still trigger this issue: 

It is due to mixing tabs and spaces in indentation before the cases, which is invisible. It makes them unaligned, thus appear to be missing.

Tabs are not allowed by Elm and it confuses the IDE plugin as well.

Error got from Elm:
```
I ran into a tab, but tabs are not allowed in Elm files.

76| 		    updateLoading msg user

    ^
Replace the tab with spaces.
```

Converting all tabs to spaces using the IDE fixed the false positives on the affected case expressions.

PR #217, opened 2019-01-12
Add test for non-applicability of case branch inspection

ajalt:
This test failed before #215, and passes now.


klazuka:
What about also adding an assertion to the constructor for `org.elm.ide.inspections.MissingCaseBranchAdder.Result.MissingVariants` which asserts that the provided map is not empty?

Issue #240, opened 2019-01-29
Fix flaky "file rename" test


klazuka:
In `org.elm.ide.refactoring.ElmRenameTest`, the test `test file rename from Data_User to Data_Quux` fails about 50% of the time in CI and 100% of the time on my machine. See failed build https://circleci.com/gh/klazuka/intellij-elm/313

The problem appears to be related to how we generate an `elm.json` file describing the test project.

klazuka:
oh, the test is still commented-out. whoops.

PR #242, opened 2019-01-31
Fix ref resolve for aliased pattern destructuring assignment

### LINKED ISSUE #207: Lexical reference error on aliased let-binding
These `let...as` constructs are syntax errors to intellij-elm but valid to the elm compiler:

intellij-elm-let-as

Thanks for the great plugin.

klazuka:
Fixes #207


klazuka:
Circle CI tests failed due to #240 again. I really need to fix that.

PR #248, opened 2019-02-03
Fix more false positives in extension record inference

ajalt:
This fixes a couple more incorrect warnings caused by some exotic typing in elm-css. I made `recordAssignable` less dry and added some comments so it's easier to reason about.

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
+++ b/src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
@@ -856,17 +856,34 @@ private class InferenceScope(
     }
 
     private fun recordAssignable(ty1: TyRecord, ty2: TyRecord): Boolean {
-        fun fieldsAssignable(t1: TyRecord, t2: TyRecord): Boolean {
-            return t1.fields.all { (k, v) -> t2.fields[k]?.let { assignable(v, it) } ?: false }
+        fun fieldsAssignable(t1: TyRecord, t2: TyRecord, strict: Boolean): Boolean {
+            return t1.fields.all { (k, v) -> t2.fields[k]?.let { assignable(v, it) } ?: !strict }
         }
-        if (ty2.isSubset) {
-            return fieldsAssignable(ty2, ty1)
+
+        // Subset record tys are created from extension record declarations or field accessor functions
+
+        return when {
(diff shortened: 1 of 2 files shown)


src/test/kotlin/org/elm/ide/inspections/TypeInferenceInspectionTest.kt
```diff
@@ -185,6 +184,30 @@ main : R
 main = foo <error descr="Type mismatch.Required: { r | x : () }Found: { y : () }">{ y = () }</error>
 """)
 
+
+    fun `test extension record argument with more fields than extension parameter`() = checkByText("""
```
ajalt: Note that these test cases are different: in the first, `Large a` is being assigned to `Small a`. In the second, `Small a` is being assigned to `Large a`.

src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
```diff
@@ -856,17 +856,34 @@ private class InferenceScope(
...
-            return fieldsAssignable(ty2, ty1)
-        }
-        val correctSize = when {
-            ty1.isSubset -> ty1.fields.size <= ty2.fields.size
-            else -> ty1.fields.size == ty2.fields.size
+
+        // Subset record tys are created from extension record declarations or field accessor functions
+
+        return when {
+            // e.g. passing an extension record argument to an extension record parameter
```
klazuka: I like how you broke this down.

The comments here reminded me of something I saw in intelli-rust. They have this thing called a [Testmark](
https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/openapiext/Testmark.kt). It allows you to check that a test is actually exercising the code-path that you want to test *and* it provides navigation between tricky logic and the test that covers the logic.

 I’m not sure whether it’s a good idea or not, but it might make sense to consider when trying to exhaustively test all of these edge cases. Or maybe it’d be better to use a less-intrusive code coverage tool?

ajalt: I don't think I've had issues that would be solved by line coverage checks. Pretty much every type inference bug is a result of a logic error or an edge case in the elm grammar I didn't know about.

PR #249, opened 2019-02-03
Fix false positive type error in referenced pattern declarations

### LINKED ISSUE #247: False positive type error in valid Elm
Elm 0.19, Elm plugin 2.2.2

The following self-contained example is marked as invalid (at the case statement), but Elm compiles it without error.

The reported error is `Type mismatch.Required: (B, D, E)Found: ((B, E), D, (B, E))

```elm
module Main exposing (main)

import Browser
import Html exposing (div)


foo :
    A
    -> B
    -> ( C, D, E )
    -> ( C, D, E )
foo a b ( c, d, e ) =
    let
        action : ActionOutcome
        action =
            PickUp (EdibleID 2)

        ( updatedB, updatedD, updatedE ) =
            case action of
                PickUp (EdibleID _) ->
                    let
                        ( b1, e1 ) =
                            ( b, e )
                    in
                    ( b1, d, e1 )

                PickUp (ExtinguisherID _) ->
                    ( b, d, e )
    in
    ( c, d, e )


type A
    = A


type B
    = B


type C
    = C


type D
    = D


type E
    = E


type ActionOutcome
    = PickUp ReferenceToPortable


type ReferenceToPortable
    = ExtinguisherID Int
    | EdibleID Int


type alias Model =
    Int


type Msg
    = Nothing


main : Program () Model Msg
main =
    Browser.sandbox
        { init = 1
        , view = \int -> div [] []
        , update = update
        }


init : Int -> ( Model, Cmd Msg )
init posixMillis =
    ( posixMillis
    , Cmd.none
    )


update : Msg -> Model -> Model
update _ model =
    model

```

Changing the first branch of the case to return `(b, d, e)` stops the false positive.

ajalt:
This was a tricky one. There were several behaviors that came together to produce this false positive.

1. Like all unannotated declarations, the `InferenceResult.ty` for pattern declarations is the ty of the declaration body. This type isn't useful for pattern declarations, but it's never used, so that's fine.
2. Except it _was_ used. This was the first bug. When inferring the type of a reference to a name bound in a pattern declaration, we returning the type of the entire declaration, rather than the bound parameter.
3. `InferenceScope` has a local cache of `resolvedDeclarations` to avoid repeating work of inferring nested declarations. We don't use the `CachedValuesManager` for nested declarations because, unlike top level declarations, they can close over bound function parameters. Caching is also important because, although inference itself is idempotent, we store its generated diagnostics, and we don't want to generate duplicate diagnostics.
4. When we infer a nested pattern declaration, we save the bound names in the `bindings` map in the current scope. This allows us to look up the bound names when they're reference. 
5. In order to bind a pattern declaration, we need to know the type of its body. This is the opposite of function declarations, in which we need to know the parameter types in order to infer the body.
6. The second bug: We used the same code path for all value declarations, checking that the body matches the annotation. This meant that for pattern declarations, the body was inferred twice. This only caused a problem because, when inferring a nested scope, `resolvedDeclarations` is copied from the parent, but `bindings` is not (since child scopes can shadow names). So looking up a bound pattern the _second_ time we inferred the body would fail. The declaration was already cached, so we didn't infer it again, but that meant that we didn't bind the pattern names.
7. These two bugs were almost always hidden by bug 3:  When `TyUnknown` was changed from an object to a class, I didn't change the last line of `beginDeclarationInference`, which was still comparing it by equality, resulting in all declarations without annotations being inferred as unknown.

Fixes #247

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
+++ b/src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
@@ -100,6 +100,14 @@ private class InferenceScope(
 
         val (declaredTy, paramCount) = bindParameters(declaration)
 
+        // The body of pattern declarations is inferred as part of parameter binding, so there's no
+        // more to do here.
+        if (declaration.pattern != null) {
+            return InferenceResult(expressionTypes, diagnostics, declaredTy)
+        }
+
+        // For function declarations, we need to infer the body and check that it matches the
+        // annotation, if there is one.
         val expr = declaration.expression
         var bodyTy: Ty = TyUnknown()
         if (expr != null && !PsiTreeUtil.hasErrorElements(expr)) {
@@ -520,7 +528,12 @@ private class InferenceScope(
                 // Pattern declarations might not have been inferred yet
                 val parentPatternDecl = parentPatternDecl(ref)
                 if (parentPatternDecl != null) {
-                    return inferReferencedValueDeclaration(parentPatternDecl)
+                    inferReferencedValueDeclaration(parentPatternDecl)
+                    // Now that the pattern's declaration has been inferred, the pattern has been
+                    // added to bindings
+                    return getBinding(ref)
+                            ?: error("failed to destructure pattern for expr of type " +
+                                    "${expr.elementType}: '${expr.text}'")
                 }
 
                 // All patterns should now be bound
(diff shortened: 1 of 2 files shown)


klazuka:
Epic bug summary.

This seems to give further support to the idea of breaking apart `ElmValueDeclaration`.

PR #250, opened 2019-02-03
Cache the ElmProject SearchScope

### LINKED ISSUE #246: Cache the SearchScope for each ElmProject in the workspace
Constructing the `GlobalSearchScope` introduced by #244 is using a lot of CPU (and probably creating a lot of garbage). It's called very often, and it should be pretty easy to cache.

Here is what the CPU profiler found typing `foo = String.length "" ++ String.fromInt 99` at the bottom of PowerLift's `Main.elm` file (using IntelliJ's hidden "retype" feature). The top 3 are regressions.

screen shot 2019-02-02 at 4 39 09 pm

klazuka:
We were constructing this `ElmProject`-specific search scope value a lot, and it was dominating the CPU profiler traces. Now we cache 2 flavors of the search scope for each Elm project: 

1. just the project files
2. the project files plus the "tests" directory as well as the "test-dependencies".

Fixes #246

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/lookup/ElmLookup.kt
+++ b/src/main/kotlin/org/elm/lang/core/lookup/ElmLookup.kt
@@ -1,9 +1,21 @@
 package org.elm.lang.core.lookup
 
 import com.intellij.openapi.diagnostic.logger
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.Key
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.psi.search.GlobalSearchScope
+import com.intellij.psi.search.GlobalSearchScopesCore
+import com.intellij.psi.util.CachedValue
+import com.intellij.psi.util.CachedValueProvider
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/elm/lang/core/lookup/ElmLookup.kt
```diff
@@ -36,4 +55,76 @@ object ElmLookup {
...
+    fun searchScopeAt(loc: ClientLocation): GlobalSearchScope {
+        val elmProject = loc.elmProject ?: return GlobalSearchScope.EMPTY_SCOPE
+
+        /*
+        Build (and cache) the search scope based on the ClientLocation.
+
+        The cache will be invalidated whenever the Elm workspace changes
+        (e.g. the user modifies an `elm.json` file)
+
+        You might be wondering why we don't just construct the search scope when
```
klazuka: @ajalt what do you think about this? Have I gone too far?

klazuka:
Without caching: 
no caching

After adding the cache, it's much improved:
after adding cache

PR #254, opened 2019-02-07
Unify type variables in function calls

### LINKED ISSUE #122: Extensible records / record types are failing in Elm 18
Sup fellas, here's an example:

image

```

iFail : Bla1 Bla2 -> String
iFail { bla2 } =
    "    ^ here I fail"


type alias Bla1 a =
    { a
        | bla : Int
    }


type alias Bla2 =
    { bla2 : Int }
```

### LINKED ISSUE #264: Code compiles, but IDE yields about type error
Got error on code that compiles fine.

```
Type mismatch. Required: Common → Dimension → Gid → (world → (EntityID, world), layer → (EntityID, layer)) → (world → (EntityID, world), layer → (EntityID, layer)) Found: a → b → c → (a → a) less... (⌘F1) 
Inspection info:The Elm type checker.
```

```elm
module Delme exposing (nonWorking)


type alias Common =
    { a : String }


type alias EntityID =
    Int


type alias Gid =
    Int


type alias Dimension =
    { width : Int, height : Int }


type alias Return world layer =
    ( world -> ( EntityID, world ), layer -> ( EntityID, layer ) )
    -> ( world -> ( EntityID, world ), layer -> ( EntityID, layer ) )


type alias Read world layer =
    { objectTile : Common -> Dimension -> Gid -> Return world layer

    --..
    }


defaultRead : Read world layer
defaultRead =
    { objectTile = \_ _ _ a -> a
    }


nonWorking =
    --TODO create other for isometric view
    let
        spec =
            { get = identity
            , set = \comps _ -> comps
            }
    in
    { spec = spec
    , read =
        { defaultRead
            | objectTile = --here is error
                \_ { height, width } _ ->
                    identity
        }
    }
```

### LINKED ISSUE #268: Function that returns lamda parses incorrectly
image

ajalt:
This PR adds a limited form of unification to function call and operator expressions. It also implements the constraints of the built-in type classes, as well as a few assorted fixes to type inference.

I removed the diagnostic that checked that a type expression argument is a record when passed to a record extension type alias. This is neither a compile time nor runtime error in Elm (although it can cause a crash in some circumstances), and complicated the `TypeReplacement` code quite a bit.

untitled

untitled2

Fixes #264 
Fixes #268 
Fixes #122

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,12 +46,8 @@ The plugin can perform type inference on arbitrary expressions in your program. 
 
 The plugin also performs type checking, marking incompatible types in red.
 
-There are 2 limitations to the type system as currently implemented:
-
-1. it does not work if a function is missing a type annotation (typically not a problem)
-2. it does not work if type variables (such as `List a` or `Maybe a` are involved)
-
-We hope to remove the type variable limitation in the future.
+The current implementation of the type system does not infer the types of parameters of unannotated
+functions, but this limitation will be removed in the future.
 
 
 # Creating a new Elm project
--- a/src/main/kotlin/org/elm/lang/core/diagnostics/ElmDiagnostic.kt
+++ b/src/main/kotlin/org/elm/lang/core/diagnostics/ElmDiagnostic.kt
@@ -116,19 +116,6 @@ class TypeMismatchError(
     }
 }
 
-class RecordBaseTypeError(
-        element: PsiElement,
-        private val actual: Ty
-) : ElmDiagnostic(element, null) {
-    override val message: String
-        get() {
-            return "Type mismatch." +
-                    "<br>Required: record" +
-                    "<br>Found: ${actual.renderedText(false, false)}"
-        }
-
-}
-
 class TypeArgumentCountError(
         element: PsiElement,
         private val actual: Int,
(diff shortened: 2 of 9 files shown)


klazuka:
🔥

klazuka:
Thank you for putting in so much work into this project 🙏 

I ran the code against some projects and it mostly looks good. In fact, the Elm standard library (things like elm/core, elm/browser, etc.) all work flawlessly. But I did run into some false positives in other projects.

Most of the false positives fall into these 2 categories:

- Json.Decode.Pipeline usage (in elm-spa-example, see `src/Article.elm`, `src/Profile.elm`, etc.) 
- function composition (in elm-spa-example see the function `tagsFromString` in `src/Page/Article/Editor.elm`)

The above are probably 95% of the errors.

There are also a few other false-positives I encountered that appear to have a different cause:

- a lambda that takes a function as a parameter:
    - see `decoderFromCred` in elm-spa-example's `src/Api.elm`
    - see `lazyStateChangeOnClick` in PowerLift's `src/JsonTree/View.elm`
- an unknown error in `viewScalar` in PowerLift's `src/JsonTree/View.elm`

A good stress test for the type checker is PowerLift's `src/App/Main.elm` which is good and nasty. There are a lot of false positives in there right now, mostly related to function composition.

ajalt:
Thanks for testing. All of those false positives should be fixed. 

1. I was inferring record constructor as the record type, and had added special cases in a couple places to account for that. I should have just inferred them as functions in the first place.
2. `>>` and `<<` actually take _three_ arguments, but the last one is always curried. I had been ignoring these operators before. I started inferring them in this PR, but I didn't partially apply them correctly.

ajalt:
Alright, everything should be working now.

PR #256, opened 2019-02-08
Improve error reporting when path to Elm compiler is misconfigured

### LINKED ISSUE #252: Plugin fails to identify global elm binary installed in nvm-versioned Node
The plugin previously worked well with system-wide node/npm, which in my case symlinked the binaries to ~/.local/bin.

When I had to switch to _nvm_, it fails to recognize _elm_ version when I point it to any of 
~/.nvm/versions/node/v10.15.1/bin (_npm_)
~/.nvm/versions/node/v10.15.1/lib/node_modules/elm/bin (_npm_) 
~/.config/yarn/global/node_modules/elm/bin (_yarn_)
~/.config/yarn/global/node_modules/.bin (_yarn_)
~/.yarn/bin (_yarn_)

Note 
I always use absolute path, _not_ ~/ prefixed path 
_elm_ command is accessible from my system PATH, and works normally. 

elm

klazuka:
Several users ran into problems where they were seeing the Elm version
reported as "0.0.0". The underlying problem was that the path to the
Elm compiler was pointing at a hash-bang wrapper script created by
npm/nvm/yarn/etc. Apparently those tools wrap native binaries with a
wrapper script, but this causes problems from the IntelliJ side.

Maybe we could do something smarter here to be able to invoke the
wrapper scripts themselves, but I suspect that it depends on shell
environment stuff which wouldn't be present inside IntelliJ.

I also added a suggested Elm binary path based on where nvm/yarn
apparently installs things.

Fixes #252

### DIFF
--- a/src/main/kotlin/org/elm/workspace/ElmToolchain.kt
+++ b/src/main/kotlin/org/elm/workspace/ElmToolchain.kt
@@ -101,22 +101,18 @@ data class ElmToolchain(val binDirPath: Path) {
     fun queryCompilerVersion(): Result<Version> {
         val elm = elmCompilerPath ?: return Result.Err("Elm compiler not found")
 
-        // NodeJS tools like npm and nvm like to play games with wrapper scripts around native binaries
-        // such as the Elm compiler. So we will try to detect the wrapper and notify the user.
-        // See https://github.com/klazuka/intellij-elm/issues/252
-        try {
-            val firstChars = ByteArray(2)
-            elm.toFile().inputStream().read(firstChars, 0, 2)
-            if (firstChars.toString(Charsets.US_ASCII) == "#!") {
-                return Result.Err("path points to a wrapper script; please use the path to the actual Elm compiler")
-            }
-        } catch (e: IOException) {
-            return Result.Err("could not peek at the first bytes in the Elm compiler")
-        }
-
         // Output of `elm --version` is a single line containing the version number (e.g. `0.19.0\n`)
         val versionString = GeneralCommandLine(elm).withParameters("--version").runExecutable()?.firstOrNull()
-                ?: return Result.Err("failed to run the Elm compiler")
+        if (versionString == null) {
+            // NodeJS tools like npm and nvm like to play games with wrapper scripts around native binaries
+            // such as the Elm compiler. So we will try to detect if the wrapper may have been the problem
+            // and notify the user. See https://github.com/klazuka/intellij-elm/issues/252
+            return if (elm.toFile().isWrapperScript()) {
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/elm/workspace/ElmToolchain.kt
```diff
@@ -96,14 +98,31 @@ data class ElmToolchain(val binDirPath: Path) {
...
+    fun queryCompilerVersion(): Result<Version> {
+        val elm = elmCompilerPath ?: return Result.Err("Elm compiler not found")
+
+        // NodeJS tools like npm and nvm like to play games with wrapper scripts around native binaries
+        // such as the Elm compiler. So we will try to detect the wrapper and notify the user.
+        // See https://github.com/klazuka/intellij-elm/issues/252
+        try {
+            val firstChars = ByteArray(2)
+            elm.toFile().inputStream().read(firstChars, 0, 2)
+            if (firstChars.toString(Charsets.US_ASCII) == "#!") {
```
klazuka: @ajalt is there a more concise way to express this?

Also, do you know what the npm wrapper scripts would look like on Windows?

ajalt: If you don't mind reading a couple of extra bytes:

```kotlin
elm.toFile().bufferedReader().readLine()?.startsWith("#!")
```

The current code doesn't close the file. If that's necessary, you could do something like:

```kotlin
elm.toFile().bufferedReader().use { it.readLine() }?.startsWith("#!")
```

or

```kotlin
elm.toFile().useLines { it }.firstOrNull()?.startsWith("#!")
```

npm on windows creates two scripts in `%APPDATA%\Roaming\npm`.

One for cygwin:

```bash
#!/bin/sh
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")

case `uname` in
    *CYGWIN*) basedir=`cygpath -w "$basedir"`;;
esac

if [ -x "$basedir/node" ]; then
  "$basedir/node"  "$basedir/node_modules/elm/bin/elm" "$@"
  ret=$?
else 
  node  "$basedir/node_modules/elm/bin/elm" "$@"
  ret=$?
fi
exit $ret
```

and one for batch:

```batch
@IF EXIST "%~dp0\node.exe" (
  "%~dp0\node.exe"  "%~dp0\node_modules\elm\bin\elm" %*
) ELSE (
  @SETLOCAL
  @SET PATHEXT=%PATHEXT:;.JS;=;%
  node  "%~dp0\node_modules\elm\bin\elm" %*
)
```

PR #261, opened 2019-02-10
elm-format gets content via stdin instead of modifying file directly.…

FloWi:
… Documents' is updated via runUndoTransparent action

### DIFF
--- a/src/main/kotlin/org/elm/ide/actions/ElmExternalFormatAction.kt
+++ b/src/main/kotlin/org/elm/ide/actions/ElmExternalFormatAction.kt
@@ -1,75 +1,71 @@
 package org.elm.ide.actions
 
 import com.intellij.execution.ExecutionException
-import com.intellij.execution.process.ProcessOutput
 import com.intellij.notification.NotificationType
 import com.intellij.openapi.actionSystem.AnAction
 import com.intellij.openapi.actionSystem.AnActionEvent
 import com.intellij.openapi.actionSystem.CommonDataKeys
+import com.intellij.openapi.editor.Document
 import com.intellij.openapi.fileEditor.FileDocumentManager
-import com.intellij.openapi.progress.ProgressManager
 import com.intellij.openapi.project.Project
-import com.intellij.openapi.vfs.VfsUtil
+import com.intellij.openapi.vfs.VirtualFile
 import org.elm.ide.notifications.showBalloon
+import org.elm.lang.core.ElmFileType
 import org.elm.lang.core.psi.ElmFile
+import org.elm.lang.core.psi.isElmFile
 import org.elm.openapiext.isUnitTestMode
-import org.elm.workspace.*
+import org.elm.workspace.ElmFormatCLI
+import org.elm.workspace.Version
+import org.elm.workspace.elmToolchain
 
 
 class ElmExternalFormatAction : AnAction() {
+
 
     override fun update(e: AnActionEvent) {
         super.update(e)
         e.presentation.isEnabled = getContext(e) != null
     }
 
-
-    private fun getContext(e: AnActionEvent): Context? {
-        val project = e.project ?: return null
-        val toolchain = project.elmToolchain ?: return null
-        val file = e.getData(CommonDataKeys.PSI_FILE) ?: return null
-        if (!file.virtualFile.isInLocalFileSystem) return null
-        if (file !is ElmFile) return null
-        val elmVersion = when (val elmProject = file.elmProject) {
-            is ElmApplicationProject -> elmProject.elmVersion
-            is ElmPackageProject -> elmProject.elmVersion.low
-            else -> return null
-        }
-        return Context(project, toolchain, file, elmVersion)
-    }
-
-
     override fun actionPerformed(e: AnActionEvent) {
-        val ctx = getContext(e) ?: return
-        val project = ctx.project
-        val virtualFile = ctx.elmFile.virtualFile
-        val elmFormat = ctx.toolchain.elmFormat
+
+        val elmFormat = e.project?.elmToolchain?.elmFormat
         if (elmFormat == null) {
-            project.showBalloon("could not find elm-format", NotificationType.ERROR)
+            e.project?.showBalloon("could not find elm-format", NotificationType.ERROR)
             return
         }
 
-        FileDocumentManager.getInstance().saveAllDocuments()
+        val ctx = getContext(e)
+
+        if (ctx == null) {
+            if (isUnitTestMode) error("something went wrong initializing the context for elm-format")
+            return
+        }
+
         try {
-            ProgressManager.getInstance().runProcessWithProgressSynchronously<ProcessOutput, ExecutionException>({
-                elmFormat.reformatFile(project, ctx.elmVersion, virtualFile)
-            }, "Running elm-format on current file...", true, project)
-            // We want to refresh file synchronously only in unit test
-            // to get new text right after `reformatFile` call
-            VfsUtil.markDirtyAndRefresh(!isUnitTestMode, true, true, virtualFile)
-        } catch (e: ExecutionException) {
-            if (isUnitTestMode) throw e
-            val message = e.message ?: "something went wrong running elm-format"
-            project.showBalloon(message, NotificationType.ERROR)
+            elmFormat.formatDocumentAndSetText(ctx.project, ctx.document, ctx.elmVersion)
+        } catch (ex: ExecutionException) {
+            if (isUnitTestMode) throw ex
+            val message = ex.message ?: "something went wrong running elm-format"
+            e.project?.showBalloon(message, NotificationType.ERROR)
+
         }
     }
 
+    private fun getContext(e: AnActionEvent): Context? {
+        val project = e.project ?: return null
+        val file = e.getData(CommonDataKeys.VIRTUAL_FILE) ?: return null
+        if (!file.isInLocalFileSystem) return null
+        if (!file.isElmFile) return null
+        val document = FileDocumentManager.getInstance().getDocument(file) ?: return null
+        val elmVersion = ElmFormatCLI.getElmVersion(project, file) ?: return null
+        return Context(project, file, document, elmVersion)
+    }
 
-    private data class Context(
+    data class Context(
             val project: Project,
-            val toolchain: ElmToolchain,
-            val elmFile: ElmFile,
+            val file: VirtualFile,
+            val document: Document,
             val elmVersion: Version
     )
 }
-
--- a/src/main/kotlin/org/elm/openapiext/Subprocesses.kt
+++ b/src/main/kotlin/org/elm/openapiext/Subprocesses.kt
@@ -35,6 +35,9 @@ import com.intellij.openapi.application.runReadAction
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.util.Disposer
 import com.intellij.util.io.systemIndependentPath
+import java.io.BufferedWriter
+import java.io.OutputStreamWriter
+import java.nio.charset.Charset
 import java.nio.file.Path
 
 private val log = Logger.getInstance("org.elm.openapiext.Subprocesses")
@@ -46,27 +49,6 @@ fun GeneralCommandLine(path: Path, vararg args: String) =
 fun GeneralCommandLine.withWorkDirectory(path: Path?) =
         withWorkDirectory(path?.systemIndependentPath)
 
-
-fun GeneralCommandLine.execute(timeoutInMilliseconds: Int? = 1000): ProcessOutput? {
-    val output = try {
-        val handler = CapturingProcessHandler(this)
-
-        log.debug("Executing `$commandLineString`")
-        if (timeoutInMilliseconds != null)
-            handler.runProcess(timeoutInMilliseconds)
-        else
-            handler.runProcess()
-    } catch (e: ExecutionException) {
-        log.warn("Failed to run executable", e)
-        return null
-    }
-
-    if (!output.isSuccess) {
-        log.warn(errorMessage(this, output))
-    }
-
-    return output
-}
 
 @Throws(ExecutionException::class)
 fun GeneralCommandLine.execute(
@@ -100,6 +82,40 @@ fun GeneralCommandLine.execute(
         handler.runProcess()
     } finally {
         Disposer.dispose(processKiller)
+    }
+
+    if (!ignoreExitCode && output.exitCode != 0) {
+        throw ExecutionException(errorMessage(this, output))
+    }
+    return output
+}
+
+@Throws(ExecutionException::class)
+fun GeneralCommandLine.execute(
+        stdIn: String? = null,
+        timeoutInMilliseconds: Int? = 2000,
+        ignoreExitCode: Boolean = false
+): ProcessOutput {
+
+    val handler = if (stdIn == null) {
+        CapturingProcessHandler(this)
+    } else {
+        val process = createProcess()
+
+        val stdInStream = process.outputStream
+        val writer = BufferedWriter(OutputStreamWriter(stdInStream))
+
+        writer.append(stdIn)
+        writer.flush()
+        writer.close()
+
+        CapturingProcessHandler(process, Charset.forName("UTF-8"), commandLineString)
+    }
+
+    val output = if(timeoutInMilliseconds == null) {
+        handler.runProcess()
+    } else {
+        handler.runProcess(timeoutInMilliseconds, true)
     }
 
     if (!ignoreExitCode && output.exitCode != 0) {
--- a/src/main/kotlin/org/elm/workspace/ElmFormatCLI.kt
+++ b/src/main/kotlin/org/elm/workspace/ElmFormatCLI.kt
@@ -1,27 +1,65 @@
 package org.elm.workspace
 
+import com.intellij.execution.ExecutionException
 import com.intellij.execution.process.ProcessOutput
-import com.intellij.openapi.Disposable
+import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.command.CommandProcessor
+import com.intellij.openapi.editor.Document
+import com.intellij.openapi.progress.ProgressManager
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.vfs.VirtualFile
+import org.elm.lang.core.psi.ElmFile
 import org.elm.openapiext.GeneralCommandLine
 import org.elm.openapiext.execute
-import org.elm.openapiext.pathAsPath
-import org.elm.openapiext.withWorkDirectory
+import org.elm.openapiext.isSuccess
+import org.elm.openapiext.toPsiFile
 import java.nio.file.Path
 
 class ElmFormatCLI(val elmFormatExecutablePath: Path) {
 
-    fun reformatFile(project: Project, elmVersion: Version, file: VirtualFile, owner: Disposable = project): ProcessOutput {
+    private fun getFormattedContentOfDocument(elmVersion: Version, document: Document): ProcessOutput {
         val arguments = listOf(
                 "--yes",
-                "--elm-version",
-                "${elmVersion.x}.${elmVersion.y}",
-                file.path
+                "--elm-version=${elmVersion.x}.${elmVersion.y}",
+                "--stdin"
         )
+
         return GeneralCommandLine(elmFormatExecutablePath)
-                .withWorkDirectory(file.parent.pathAsPath)
                 .withParameters(arguments)
-                .execute(owner)
+                .execute(document.text)
+    }
+
+    fun formatDocumentAndSetText(project: Project, document: Document, version: Version) {
+
+        val result = ProgressManager.getInstance().runProcessWithProgressSynchronously<ProcessOutput, ExecutionException>({
+            getFormattedContentOfDocument(version, document)
+        }, "Running elm-format on current file...", true, project)
+
+        if (result.isSuccess) {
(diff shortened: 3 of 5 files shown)


klazuka:
Thanks! I'll review this today or tomorrow.

src/main/kotlin/org/elm/ide/actions/ElmExternalFormatAction.kt
```diff
@@ -27,14 +29,10 @@ class ElmExternalFormatAction : AnAction() {
...
-        val file = e.getData(CommonDataKeys.PSI_FILE) ?: return null
+        val file: PsiFile = e.getData(CommonDataKeys.PSI_FILE) ?: return null
         if (!file.virtualFile.isInLocalFileSystem) return null
         if (file !is ElmFile) return null
-        val elmVersion = when (val elmProject = file.elmProject) {
-            is ElmApplicationProject -> elmProject.elmVersion
-            is ElmPackageProject -> elmProject.elmVersion.low
-            else -> return null
-        }
+        val elmVersion = getElmVersion(file) ?: return null
```
klazuka: Your change no longer uses the `elmVersion` from the `Context` object. So you can remove this line entirely. That will leave `getElmVersion()` with only a single caller, and it should probably be inlined.

src/main/kotlin/org/elm/ide/actions/ElmExternalFormatAction.kt
```diff
@@ -49,14 +47,9 @@ class ElmExternalFormatAction : AnAction() {
...
 
-        FileDocumentManager.getInstance().saveAllDocuments()
         try {
-            ProgressManager.getInstance().runProcessWithProgressSynchronously<ProcessOutput, ExecutionException>({
-                elmFormat.reformatFile(project, ctx.elmVersion, virtualFile)
-            }, "Running elm-format on current file...", true, project)
-            // We want to refresh file synchronously only in unit test
-            // to get new text right after `reformatFile` call
-            VfsUtil.markDirtyAndRefresh(!isUnitTestMode, true, true, virtualFile)
+            val document = FileDocumentManager.getInstance().getDocument(virtualFile)!!
```
klazuka: The point of the `Context` object is to gather together everything that the action needs to successfully run so that you don't have to `!!` things. So I would either add the document to the context or do a check and early return at the beginning of `actionPerformed`.

src/main/kotlin/org/elm/workspace/ElmFormatUtils.kt
```diff
@@ -0,0 +1,69 @@
...
+                            }
+                }
+            }
+        }
+    }
+
+    fun projectForFile(file: VirtualFile): Project? {
+        return ProjectManager.getInstance().openProjects.find { project ->
+            val fileIndex = ProjectRootManager.getInstance(project).fileIndex
+            // FIXME source-directories listed in elm.json aren't marked as Source Root in the IDEA project
```
klazuka: Yeah, the 'source root' thing was never implemented. But just today I started working on some stuff related to that (#262). It doesn't actually do the work of marking src/test/excluded roots, but that will be coming later.

src/main/kotlin/org/elm/workspace/ElmFormatUtils.kt
```diff
@@ -0,0 +1,69 @@
...
+                            .getInstance()
+                            .runUndoTransparentAction {
+                                document.setText(formatted)
+                            }
+                }
+            }
+        }
+    }
+
+    fun projectForFile(file: VirtualFile): Project? {
```
klazuka: Not used: please remove.

src/main/kotlin/org/elm/workspace/ElmFormatUtils.kt
```diff
@@ -0,0 +1,69 @@
...
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.project.ProjectManager
+import com.intellij.openapi.roots.ProjectRootManager
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.psi.PsiFile
+import org.elm.lang.core.psi.ElmFile
+import org.elm.openapiext.isSuccess
+import org.elm.openapiext.toPsiFile
+
+object ElmFormatUtils {
```
klazuka: All of the code in this file should be moved into `ElmFormatCLI`. Either as helper functions outside of the class or as methods on the companion object.

src/main/kotlin/org/elm/workspace/ElmFormatUtils.kt
```diff
@@ -0,0 +1,69 @@
...
+            else -> return null
+        }
+
+        return elmVersion
+    }
+
+    fun formatDocumentAndSetText(project: Project, document: Document, elmFormatCLI: ElmFormatCLI) {
+
+        val file = FileDocumentManager.getInstance().getFile(document)
+        val version = getElmVersion(project, file!!)
```
klazuka: Instead of doing `file!!` (and `version!!` a few lines down), just pass the data to the function.

src/main/kotlin/org/elm/workspace/ElmFormatUtils.kt
```diff
@@ -0,0 +1,69 @@
...
+
+        if (result.isSuccess) {
+            val formatted = result.stdout
+            val source = document.text
+
+            if (source != formatted) {
+                ApplicationManager.getApplication().runWriteAction {
+                    CommandProcessor
+                            .getInstance()
+                            .runUndoTransparentAction {
```
klazuka: I can see why making the action "undo transparent" will make sense when it's part of an on-save hook, but right now it's an explicitly invoked user action, and it seems better if it had its own, separate undo step.

src/main/kotlin/org/elm/openapiext/Subprocesses.kt
```diff
@@ -108,6 +112,60 @@ fun GeneralCommandLine.execute(
...
 }
 
+@Throws(ExecutionException::class)
+fun GeneralCommandLine.executeWithStdIn(
+        owner: Disposable,
+        stdIn: String,
+        ignoreExitCode: Boolean = false
+): ProcessOutput {
+
+    val process = this.createProcess()
```
klazuka: remove `this.` from here and also line 125.

src/main/kotlin/org/elm/openapiext/Subprocesses.kt
```diff
@@ -108,6 +112,60 @@ fun GeneralCommandLine.execute(
     return output
 }
 
+@Throws(ExecutionException::class)
+fun GeneralCommandLine.executeWithStdIn(
```
klazuka: Maybe it would be better to instead extend the `execute` function that takes a timeout? We don't want elm-format to run indefinitely, and the timeout case is easier than all this `Disposer` junk.

I would set the timeout to be 2 seconds. That should be plenty.

FloWi:
Lots of changes to make, but a good way for me to learn kotlin, since it's only my 2nd day of usage. So thanks for the good review!

I'll make the changes later today.

src/main/kotlin/org/elm/openapiext/Subprocesses.kt
FloWi: Good idea. I removed my implementation and adjusted the existing one to take an optional stdin. But please take a look at the timeout in org.elm.workspace.ElmCLI#installDeps.

src/main/kotlin/org/elm/workspace/ElmFormatUtils.kt
FloWi: Ok, I adjusted that. As far as I understand the docs the "outermost" action gets pushed to the undo-stack including its sub-actions as one action.

src/main/kotlin/org/elm/workspace/ElmFormatUtils.kt
FloWi: nice!

src/main/kotlin/org/elm/ide/actions/ElmExternalFormatAction.kt
FloWi: Good idea! I adjusted the `Context` object and pass it to the formatting method. Will use it also in my SaveHook impl.

Didn't look at it properly beforehand and thought it was something IDEA-internal.

FloWi:
OK, I'm done with my changes. Don't know exactly why CI breaks. Looks like elm-format isn't installed, right? Please take a look at my refactoring of `GeneralCommandLine.execute` and esp. the callers at `ElmCLI.make` and `ElmCLI.installDeps` regarding the timeouts.

klazuka:
No problem. I don’t mind doing code review. I’m excited about the work that you’re doing: the elm-format integration has always been an afterthought, but I should I have spent more time on it earlier.

FloWi:
You did a great job creating this plugin and it continually improved it over time. It has a lot of features - no wonder you didn't complete everything in the beginning. I'm glad I could help out and learned a bit about IDEA plugins in doing so :)

FloWi:
Don't merge yet. The undo stuff is broken. 
image

Added `CommandProcessor.getInstance().runUndoTransparentAction` back in, but now it says "Undo Typing"... Need to debug this.

FloWi:
Ok, undo works again. I'll squash all my commits into one, ok? Otherwise the git log looks crazy.

klazuka:
Thanks. 

There is one flaky test that frequently causes CI failure (related to rename refactoring). But in your case, the failures that I saw seemed to be related to the Elm package repo being unavailable. 

I don’t have elm-format currently installed in CI, so, in principle, your code would not have been involved in any CI failures.

FloWi:
Oh, I see. Flaky tests are evil. I'll work on the savehook tomorrow. Going climbing now and probably won't have much fun typing after that :)

It works so far already but I still need to make it configurable. And also check if a file is an actual elm-file - currently it formats `*.*`...

src/main/kotlin/org/elm/openapiext/Subprocesses.kt
```diff
@@ -70,36 +71,32 @@ fun GeneralCommandLine.execute(timeoutInMilliseconds: Int? = 1000): ProcessOutpu
 
 @Throws(ExecutionException::class)
 fun GeneralCommandLine.execute(
-        owner: Disposable,
```
klazuka: Whoops, this was probably my fault for not explaining things well. The disposer code path needs to stick around. That's what `ElmCLI` calls when compiling a project (part of the CI setup and a (currently hidden) action to compile an Elm file).

Basically, there should be 2 core execution APIs:
- timeout-based, in which case no Disposer is needed (I think)
- disposer-based (can run indefinitely or until IntelliJ is closed)

The special handling for piping in stdin should be done with the timeout-based codepath since the user doesn't want to wait indefinitely (and because the code is simpler).

src/main/kotlin/org/elm/workspace/ElmFormatCLI.kt
```diff
@@ -1,27 +1,82 @@
...
         return GeneralCommandLine(elmFormatExecutablePath)
-                .withWorkDirectory(file.parent.pathAsPath)
                 .withParameters(arguments)
-                .execute(owner)
+                .execute(document.text)
     }
+
+    companion object {
+
+        fun getElmVersion(project: Project, file: VirtualFile): Version? {
```
klazuka: this should be inlined into the caller.

src/main/kotlin/org/elm/workspace/ElmFormatCLI.kt
```diff
@@ -1,27 +1,82 @@
...
+                is ElmPackageProject -> elmProject.elmVersion.low
+                else -> return null
+            }
+
+            return elmVersion
+        }
+
+        fun formatDocumentAndSetText(ctx: Context) {
+
+            val file = FileDocumentManager.getInstance().getFile(ctx.document)
```
klazuka: What's the point of getting the VirtualFile here? Isn't it in the context object?

src/main/kotlin/org/elm/workspace/ElmFormatCLI.kt
```diff
@@ -1,27 +1,82 @@
...
+            val version = getElmVersion(ctx.project, file!!)
```
klazuka: This is in the context, too.

src/main/kotlin/org/elm/workspace/ElmFormatCLI.kt
```diff
@@ -1,27 +1,82 @@
...
+            val elmVersion: Version = when (val elmProject = psiFile.elmProject) {
+                is ElmApplicationProject -> elmProject.elmVersion
+                is ElmPackageProject -> elmProject.elmVersion.low
+                else -> return null
+            }
+
+            return elmVersion
+        }
+
+        fun formatDocumentAndSetText(ctx: Context) {
```
klazuka: this should probably become an instance method on `ElmFormatCLI`. Then you can remove the CLI object from the Context class.

src/main/kotlin/org/elm/workspace/ElmCLI.kt
```diff
@@ -11,22 +10,22 @@ class ElmCLI(private val elmExecutablePath: Path) {
 
     // TODO [kl] allow the caller to specify the main entry point Elm file
 
-    fun make(owner: Disposable, elmProject: ElmProject): ProcessOutput {
```
klazuka: This parameter will need to be added back in once the `execute` method that takes a Disposable is restored.

src/main/kotlin/org/elm/workspace/ElmCLI.kt
```diff
@@ -11,22 +10,22 @@ class ElmCLI(private val elmExecutablePath: Path) {
...
+    fun make(elmProject: ElmProject): ProcessOutput {
         val workDir = elmProject.manifestPath.parent
         return GeneralCommandLine(elmExecutablePath)
                 .withWorkDirectory(workDir)
                 .withParameters("make", "src/Main.elm", "--output=/dev/null")
-                .execute(owner, ignoreExitCode = true)
+                .execute(ignoreExitCode = true)
     }
 
-    fun installDeps(owner: Disposable, elmProjectManifestPath: Path): ProcessOutput {
```
klazuka: This needs to be changed back too.

src/main/kotlin/org/elm/workspace/ElmFormatCLI.kt
```diff
@@ -1,27 +1,82 @@
...
+    companion object {
+
+        fun getElmVersion(project: Project, file: VirtualFile): Version? {
+            val psiFile = file.toPsiFile(project) ?: return null
+            return getElmVersion(psiFile)
+        }
+
+        fun getElmVersion(psiFile: PsiFile): Version? {
+            if (psiFile !is ElmFile) return null
+            val elmVersion: Version = when (val elmProject = psiFile.elmProject) {
```
klazuka: return directly here. no need to create a local variable.

src/main/kotlin/org/elm/workspace/ElmFormatCLI.kt
```diff
@@ -1,27 +1,82 @@
...
+                        CommandProcessor.getInstance().runUndoTransparentAction {
+                            ctx.document.setText(formatted)
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    data class Context(
```
klazuka: The Context class should be moved back into ElmExternalFormatAction. It’s kind of a UI concern. It exists solely so that the logic for determining whether the action should be enabled or not is the same as the logic for acquiring the data that is necessary to perform the formatting.

FloWi:
Ok, I'm done with my changes. Hope, everything's cleaner now.

FloWi:
Now I finally managed to use git properly :)

src/main/kotlin/org/elm/workspace/ElmFormatCLI.kt
```diff
@@ -1,27 +1,67 @@
...
+        val result = ProgressManager.getInstance().runProcessWithProgressSynchronously<ProcessOutput, ExecutionException>({
+            getFormattedContentOfDocument(version, document)
+        }, "Running elm-format on current file...", true, project)
+
+        if (result.isSuccess) {
+            val formatted = result.stdout
+            val source = document.text
+
+            if (source != formatted) {
+                ApplicationManager.getApplication().runWriteAction {
```
klazuka: While we will want the undo-transparent action later when this is invoked via save hook, right now we want to add to the undo stack. I know you said that you tried and ran into problems. I gave it a go and had problems too, but I eventually figured it out. You have to call `CommandProcessor#executeCommand` and do the document update inside the runnable that you give it.

Here's the solution that works:
```kotlin
        if (result.isSuccess) {
            val formatted = result.stdout
            if (document.text != formatted) {
                CommandProcessor.getInstance().executeCommand(project, {
                    ApplicationManager.getApplication().runWriteAction {
                        document.setText(formatted)
                    }
                }, "apply elm-format output", null, document)
            }
        }
```

I also made it so that we wait to acquire the write action until as late as possible--this follows what I've seen in other usages of `executeCommand` from within the IntelliJ codebase.

src/main/kotlin/org/elm/ide/actions/ElmExternalFormatAction.kt
```diff
@@ -1,75 +1,69 @@
...
+            if (isUnitTestMode) throw ex
+            val message = ex.message ?: "something went wrong running elm-format"
+            e.project?.showBalloon(message, NotificationType.ERROR)
+
         }
     }
 
+    private fun getContext(e: AnActionEvent): Context? {
+        val project = e.project ?: return null
+        val file = e.getData(CommonDataKeys.PSI_FILE) ?: return null
```
klazuka: Now that we're using the VirtualFile, we don't actually need the PsiFile here at all. To get the virtual file: 
`val file = e.getData(CommonDataKeys.VIRTUAL_FILE) ?: return null`

Then to check that it's an Elm file, do `if (file.fileType != ElmFileType) return null`.

src/main/kotlin/org/elm/openapiext/Subprocesses.kt
```diff
@@ -108,6 +111,40 @@ fun GeneralCommandLine.execute(
     return output
 }
 
+@Throws(ExecutionException::class)
+fun GeneralCommandLine.execute(
```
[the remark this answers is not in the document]
FloWi: Thx for noticing since I forgot to ask you, why this is not a compile-error in kotlin :)

src/main/kotlin/org/elm/workspace/ElmFormatCLI.kt
FloWi: Oh, how cool that you figured it out. 👍 I wish Jetbrains plugin documentation would be a bit better.

src/main/kotlin/org/elm/workspace/ElmFormatCLI.kt
```diff
@@ -1,27 +1,67 @@
...
+                }
+            }
+        }
+    }
+
+
+    companion object {
+
+        fun getElmVersion(project: Project, file: VirtualFile): Version? {
+            val psiFile = file.toPsiFile(project) ?: return null
```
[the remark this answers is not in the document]
FloWi: magic syntax :)

src/main/kotlin/org/elm/workspace/ElmFormatCLI.kt
FloWi: I tried the new behaviour with my current impl of the save hook. It works as expected and there is no "undo reload from disk". :)

klazuka:
@FloWi nice. I added a few commits to finish things off. I decided to remove the default key-binding until we have time to look into our options here.

FloWi:
Cool, thx for the fixes! 

Just fyi: I created a build of the plugin with the default keybinding and manually installed it into my IntelliJ with the existing scalafmt plugin. Both actions work fine - and our impl is better, because we properly disable the action, when we're not in an elm file :)

What other options do you want to look into?

klazuka:
The other options are:
1. allow the user to define a key-binding in the Elm settings UI
2. automatically assign the same key-binding that IntelliJ uses to re-format code (which on my Mac is Cmd+Option+L, but it varies with platform).

Option 2 seems nice to me (at least until I actually provide a proper, in-process whitespace formatter), but there are questions around it like "is it even possible?" and "is the key-binding different on other platforms and, if so, how to handle that?".

FloWi:
Option 2 would be the best. But I'd guess you can't override the standard idea formatter by reassigning the keybinding. 

You could try implementing your own formatter and use elm-format under the hood. Might be problematic, because IDEA allows formatting of selection of a document. 

Are you trying to reimplement elm-format in your plugin in the long run?

klazuka:
> Are you trying to reimplement elm-format in your plugin in the long run?

Ideally no. I don't have the time, and it seems like it would be a moving target. They keep making changes to the formatting rules. But it would be the "right" thing to do from an IntelliJ perspective. There were some messages on the IntelliJ dev forums that discouraged plugin authors from invoking external formatters since they want it to be automatic and occur as the user presses the Enter key, etc.

FloWi:
Something just came to mind. 

If elm-format would only do whitespace rearrangement would it be possible to diff unformatted vs. formatted and manipulate the psiStuff to match the formatted version? Similar to the shadow dom. 

Just an idea with no knowledge of how deep that rabbithole would be...

klazuka:
you could do that, but you'd still have invoke an external process, and I don't think you want to be blocking when, say, the user presses the "Return" key. We'd have to make it run async, at which point you should just use the output from elm-format directly (as we are doing now).

FloWi:
Right, that sounds very wrong.

PR #263, opened 2019-02-11
Be more careful about lookup when module name is empty

### LINKED ISSUE #260: Incorrectly flagged case statements
The first `case` statement in each `.elm` file of the tutorial is flagged, "Case expression is not exhaustive". This happens in the Elm plugin v2.2.3 and v2.2.2. I'm using the latest versions of IntelliJ IDEA Ultimate 2018.3.4 (build #U-183.5429.30), Elm 0.19.0 (via [MacOS installer](https://github.com/elm/compiler/releases/download/0.19.0/installer-for-mac.pkg)), and node v11.9.0.

To verify that the bug did not depend upon my particular Elm project, I created a new project from Evan's [elm-architecture-tutorial](https://github.com/evancz/elm-architecture-tutorial). Here's a screenshot of file [`examples/04-maybe.elm`](https://github.com/evancz/elm-architecture-tutorial/blob/master/examples/04-maybe.elm) in IntelliJ with the Elm plugin:

screen shot 2019-02-10 at 1 51 39 pm
When you mouse over the flagged `case`, you can see the error message:
screen shot 2019-02-10 at 1 48 31 pm

(I've separated this issue from [this one in elm-external-tools](https://github.com/jlagerweij/elm-external-tools/issues/8), because it arrises even when Elm External Tools is not installed.)

klazuka:
Fixes #260

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/lookup/ElmLookup.kt
+++ b/src/main/kotlin/org/elm/lang/core/lookup/ElmLookup.kt
@@ -46,15 +46,24 @@ object ElmLookup {
                 .filterIsInstance<T>()
     }
 
-
-    /** Find the named element with [name] declared in [module] and visible to [clientLocation] */
+    /**
+     * Find the named element with [name] in [module] which is visible to [clientLocation].
+     *
+     * Q) Why does this function return a list? Shouldn't module + name uniquely identify
+     *    an entity in Elm?
+     *
+     * A) No, it's not sufficient. Depending on how narrowly you restrict the return type,
+     *    your query could match things in both the value and type namespaces. Plus, a user
+     *    might have multiple files in their project each with the same module name. Sure,
+     *    their app won't compile if they try to import the module, but we still need
+     *    to deal with it.
+     */
     inline fun <reified T : ElmNamedElement> findByNameAndModule(
             name: String,
             module: String,
             clientLocation: ClientLocation
-    ): T? =
-            findByName<T>(name, clientLocation)
-                    .find { it.moduleName == module }
+    ): List<T> =
+            findByName<T>(name, clientLocation).filter { it.moduleName == module }
 
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/elm/lang/core/lookup/ElmLookup.kt
```diff
@@ -52,9 +52,24 @@ object ElmLookup {
...
+                /*
+                The user is looking for something that belongs to an unnamed module. Generally
+                this would only happen in a toy Elm project where you haven't yet given your module
+                a name. But it is possible that a user might have an entire directory of such Elm files,
+                each without a name (see https://github.com/klazuka/intellij-elm/issues/260).
+
+                There's only one sensible way to handle it: consider it a match iff both the candidate
+                and the caller belong to the same file.
+                */
+                results.find { it.elmFile == clientLocation }
```
klazuka: This comparison against `clientLocation` is kinda dirty since there may be non-`ElmFile` implementations of the `ClientLocation` interface. But the only reason that interface exists at all is to make it easier to test: 99% of the time, it will be an `ElmFile`.

ajalt:
I think that this is a good enough solution, although it's not a total fix. The lookup problem isn't specific to empty module names. If multiple files in the same project delcare the same module name, the issue still exists.

For example:

```kotlin
    fun `test multiple files with same module and type name`() = checkFixIsUnavailableByFileTree("Add missing case branches", """
--@ File1.elm
module Main exposing (Msg)
type Msg = Foo
--@ File2.elm
module Main exposing (Msg)
type Msg = Bar
foo : Msg -> ()
foo it =
    case{-caret-} it of
        Bar ->
            ()
--@ File3.elm
module Main exposing (Msg)
type Msg = Qux
""")
```

However, we don't necessarily need to worry about that.

klazuka:
Oh, good point. I think there’s a simple solution: regardless of whether the module name is empty, if we have more than one result with the same module name, always prefer the one which belongs to the same file as the client/caller (right now we are arbitrarily taking the first element). 

There would be cases where it wouldn’t work (especially when the query’s reified result type is overly general), but it’s probably good enough.

klazuka:
@ajalt what do you think? I decided to junk-up the caller. I didn't like the idea of hiding bugs in `ElmLookup`. The caller is in a much better place to resolve the ambiguity.

klazuka:
Uggh. Rename test failed again. I'm just going to disable that test until I knuckle-down and fix it (#240)

PR #283, opened 2019-02-15
elm-format action and on-file-save disabled on files with errors

FloWi:


src/main/kotlin/org/elm/ide/actions/ElmExternalFormatAction.kt
```diff
@@ -58,6 +60,8 @@ class ElmExternalFormatAction : AnAction() {
         if (!file.isInLocalFileSystem) return null
         if (!file.isElmFile) return null
         val document = FileDocumentManager.getInstance().getDocument(file) ?: return null
+        val psiFile = PsiDocumentManager.getInstance(project).getPsiFile(document) ?: return null
+        if (PsiTreeUtil.hasErrorElements(psiFile)) return null
```
klazuka: In this case where the user has explicitly invoked the action (e.g. by keybinding), we should show a message to the user saying that elm-format cannot be run until all parse errors are fixed. Silent is ok only when triggered by the save hook.

I'm not sure what all of the options are for displaying a non-intrusive message to the user, but there are some "balloon" notifications that we use (see org.elm.ide.notifications.UtilsKt#showBalloon).
Ideally it would be something that dismisses automatically after a few seconds.

FloWi: Of course, you're absolutely right. 

Haven't worked on user interfaces for ages and this would actually be a very bad UX - like one where you press the keys again and a bit harder in case they didn't work at the first try. 

I'll optimize that on the weekend.

klazuka:
@FloWi what's your status on this? I'd like to get everything related to elm-format done this week. If you don't have time, then I'd be fine with taking over to finish the work. Just let me know either way.

FloWi:
@klazuka I'm working on it now. Real life collided with my virtual plans on the weekend :)

FloWi:
I'm displaying an error-balloon now as you suggested. Also found a way to test it.

klazuka:
Nice! I just added a commit that simplifies the test by using an adapter class to avoid some boilerplate. Your functionality worked perfectly. Thanks!

Issue #274, opened 2019-02-13
GLSL shader lex/parse error when contains `||`


klazuka:
```
fragmentShader : Shader {} Model { vcoord : Vec2 }
fragmentShader =
    [glsl|
        void main () {
            if (true || false) {}
        }
    |]
```

I'm pretty sure that this is just a silly mistake in the lexer finding the end delimiter.

PR #287, opened 2019-02-16
Fix lexer error in GLSL expressions

ajalt:
The lexer rules previously didn't allow `|` characters anywhere in the content of the shader. 

I added basic lexer and parser tests for shaders. They aren't extensive, but until we do language injection they're probably good enough.


klazuka:
Fixes #274 

Thanks!

PR #289, opened 2019-02-17
Add UI to change the keyboard binding for elm-format

klazuka:
The Elm settings screen now shows the current keybinding for running `elm-format` as well as a hyperlink to change the shortcut via IntelliJ's "Keymap" screen.

I also re-organized the Elm settings screen so that it's a little nicer.

screen shot 2019-02-16 at 6 03 10 pm


klazuka:
@FloWi FYI

FloWi:
That looks pretty straightforward. 

Btw: I really like the redesign of the settings ui! It looks almost exactly like my draft that I tried to build with the swing-designer - but the code is way cleaner!

PR #291, opened 2019-02-17
Fix "cannot share data context between Swing events" bug

klazuka:
Using the `project` accessor on `AnActionEvent` uses `DataContext`
under-the-hood, which represents input focus. If the input focus changes,
and you try to use the old data context, IntelliJ will throw an exception.

A JetBrains engineer explains it here:
https://intellij-support.jetbrains.com/hc/en-us/community/posts/207379815/comments/207186065

And they recommend that you collect all of the data you need up-front
(at the beginning of `actionPerformed`). So that's what we do now.


klazuka:
FYI @FloWi this may affect what you're doing (showing a nice error when the user tries to run elm-format but the source code has syntax errors).

PR #300, opened 2019-02-18
Remove reserved keywords from lexer

### LINKED ISSUE #297: A type alias with a field called export is shown as error
Any type alias with keywords as fields seem to cause errors.

image

I tried this with words like type or import and also got an error.

ajalt:
As of Elm 0.19, it appears that there are no longer any reserved keywords in the compiler.

https://github.com/elm/compiler/blob/d09b95aac107291d50e8ee391d3288fc18c3891a/compiler/src/Parse/Keyword.hs#L29
https://ellie-app.com/4LRctgzwFJGa1

Fixes #297


ajalt:
I did write a [test for this change](https://github.com/ajalt/intellij-elm/commit/3b078f6b413bea080ae912144364856600e7cf48), but given that it's very unlikely that we'll have a regression on this, I chose not to add it as part of the PR. I can do so if you'd like.

klazuka:
Thanks! No test necessary.

PR #304, opened 2019-02-19
added elm-format version to settings screen

FloWi:
Here's a draft of the elm-format version label in the settings ui. Works fine but feels like too much red if elm-format isn't found at the path. What do you think?


klazuka:
There is a lot of red flashing and text changing, but it's ok for now. I will do some UI work to make it nicer, especially since we will also be adding an elm-test integration down the road, and that, too, will need the same kind of text-field + version/status treatment.

PR #305, opened 2019-02-21
Add GLSL language injection

ajalt:
IntelliJ allows you to inject languages from other plugins if they register a mimetype. This PR adds language injection to GLSL code blocks if the user has a plugin that supports it. If no plugin is found, GLSL blocks will behave the same way they currently do.

I also added mimetypes to the elm language registration so that other plugins can inject elm code if they want to.

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/psi/elements/ElmGlslCodeExpr.kt
+++ b/src/main/kotlin/org/elm/lang/core/psi/elements/ElmGlslCodeExpr.kt
@@ -1,8 +1,27 @@
 package org.elm.lang.core.psi.elements
 
 import com.intellij.lang.ASTNode
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiLanguageInjectionHost
+import org.elm.ide.injection.GlslEscaper
 import org.elm.lang.core.psi.ElmAtomTag
 import org.elm.lang.core.psi.ElmPsiElementImpl
+import org.elm.lang.core.psi.ElmPsiFactory
+import org.elm.lang.core.psi.ElmTypes
 
 
-class ElmGlslCodeExpr(node: ASTNode) : ElmPsiElementImpl(node), ElmAtomTag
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/elm/lang/core/psi/elements/ElmGlslCodeExpr.kt
```diff
@@ -1,8 +1,27 @@
...
 import org.elm.lang.core.psi.ElmPsiElementImpl
+import org.elm.lang.core.psi.ElmPsiFactory
+import org.elm.lang.core.psi.ElmTypes
+
+
+class ElmGlslCodeExpr(node: ASTNode) : ElmPsiElementImpl(node), ElmAtomTag, PsiLanguageInjectionHost {
+    val content: List<PsiElement> get() = findChildrenByType(ElmTypes.GLSL_CODE_CONTENT)
+
+    override fun updateText(text: String): PsiLanguageInjectionHost {
+        val expr = ElmPsiFactory(project).createGlslExpr(text)
```
ajalt: I'm curious if there's a more efficient way to do this. given the way we lex these blocks, I couldn't think of one.

klazuka:
Do you know if there's a way to get IntelliJ to display the "hey, there's a plugin that you can install to display this language" banner to appear? I know it does it for regular files, but maybe you can do something special with injected languages?

src/main/kotlin/org/elm/lang/core/psi/elements/ElmGlslCodeExpr.kt
klazuka: No idea. This seems fine for now.

klazuka:
Attaching a screenshot for bystanders. This shows Elm code with embedded GLSL:
screen shot 2019-02-21 at 9 43 11 am

justgook:
btw is there way to disable that green background ?

PR #310, opened 2019-02-23
produce elm test run configs in project context menu

frawa:
and as side effect, suggest config names while choosing elm folder

### DIFF
--- a/src/main/java/org/frawa/elmtest/core/ElmProjectTestsHelper.java
+++ b/src/main/java/org/frawa/elmtest/core/ElmProjectTestsHelper.java
@@ -19,6 +19,10 @@ public class ElmProjectTestsHelper {
         workspaceService = ServiceManager.getService(project, ElmWorkspaceService.class);
     }
 
+    public static boolean isElmProject(String candidate, Project project) {
+        return new ElmProjectTestsHelper(project).nameByProjectDirPath(candidate).isPresent();
+    }
+
     public Stream<String> allNames() {
         return getTestableProjects()
                 .map(ElmProject::getPresentableName);
--- a/src/main/java/org/frawa/elmtest/core/ElmProjectTestsHelper.java
+++ b/src/main/java/org/frawa/elmtest/core/ElmProjectTestsHelper.java
@@ -19,10 +19,6 @@ public class ElmProjectTestsHelper {
         workspaceService = ServiceManager.getService(project, ElmWorkspaceService.class);
     }
 
-    public static boolean isElmProject(String candidate, Project project) {
-        return new ElmProjectTestsHelper(project).nameByProjectDirPath(candidate).isPresent();
-    }
-
     public Stream<String> allNames() {
         return getTestableProjects()
                 .map(ElmProject::getPresentableName);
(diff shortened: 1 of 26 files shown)


klazuka:
Cool! I was just thinking today that we will want to surface the `elm-test` integration better before we release this feature. I didn't realize it was so easy to do contextually based on the current file/Psi.

When I try to run it on [elm-json-tree-view](https://github.com/Microsoft/elm-json-tree-view), it doesn't suggest any run configs. I ran it in the debugger and noticed that `org.frawa.elmtest.core.ElmProjectTestsHelper#nameByProjectDirPath` was being passed the value `"/Users/keith/dev/elm-json-tree-view/.idea"`. So there's something wrong about the logic for determining the path because it shouldn't be looking at the `.idea` directory.

If I understand what your code is trying to do, you might want to instead use `org.elm.workspace.ElmWorkspaceService#findProjectForFile`. This is what I use to determine the associated `ElmProject` (`elm.json`) for any ElmPsiElement. It should work for you, and then you might be able to get rid of some of your lookup functions, too.

klazuka:
In addition to surfacing the test integration through this extension point, I was also thinking that it'd be nice to do via a line marker in the gutter. But that would sortof imply that it's possible to run test suites individually, which, if I understand correctly, the `elm-test` runner does not support.

What are your thoughts about the line marker? Have you tried it and ruled it out?

klazuka:
@frawa this is the last thing remaining before I publish the big release (elm-test support, elm-format support, GLSL support, and aliased-import quick-fix). 

Is there anything you need from me? I haven't tested it yet because I wasn't sure if you were finished, but it looks like your most recent commit (8cbe6c4) will only work when the Elm project (`elm.json`) is located directly in the content root of the IntelliJ project. But what if I had a project with both a directory for a backend server and another directory for the frontend code? It seems like that wouldn't work.

frawa:
The current code works for both project layouts, with and without modules. (At least that's what I thought to see with my manual testing.)

I am about to add tests to `ElmProjectHelperTest` to cover both those project layouts, but did not succeed to create a fixture with modules inside. (Do you have hints on that part?) Yesterday night I ran out of time, let's see what today brings.

I preferred to push without those tests, to unblock more testing towards release.

frawa:
On a side node, for now I prefer to stick closer to IntelliJ's project/module model, and not `org.elm.workspace.ElmWorkspaceService#findProjectForFile` as you suggested.

I got the feeling there a ways to avoid asking users to maintain a separate list of elm projects ... to be explored.

klazuka:
> I got the feeling there a ways to avoid asking users to maintain a separate list of elm projects ... to be explored.

Quite possibly. IntelliJ's project/module/library system seemed complicated and somewhat Java-specific. The Elm plugin is largely based off of [intellij-rust](https://github.com/intellij-rust/intellij-rust). Rust uses `cargo.toml` files which are roughly equivalent to `elm.json` files. So I followed their example and built the project management stuff off-to-the-side. I assumed that the Rust team had reasons for not using IntelliJ's project/module/library system, but maybe that was a mistake.

If you'd like to dive into it more and report your findings, I'd definitely be interested.

klazuka:
When I tested it on an IntelliJ project with nested structure like:
```
root
  - server
    - src
  - client
    - src
    - tests
    - elm.json
```

The contextual suggestion would not work inside `root/client/tests`.

So I just pushed a commit to your branch that uses the `ElmWorkspaceService` to locate the root of the containing `ElmProject`.

I'm going to merge this now and prepare a release candidate build this morning.

PR #311, opened 2019-02-24
Allow user to configure location of `elm-format` and `elm-test` executables

klazuka:
I also refactored `ElmToolchain` and related classes.

Notable Changes:
- the user can now set the path to `elm-format` and `elm-test`
- added an "Auto Discover" button to the settings UI to find each tool based on common locations
- the path to `elm-format` and `elm-test` are now persisted in workspace settings
- code related to suggesting program locations has been moved into its own class, `ElmSuggest`
- created an `ElmTestCLI` class to match the existing `ElmCLI` and `ElmFormatCLI` classes
- the toolchain on the ElmWorkspace is no longer nullable

I'm still not 100% happy with how the `Settings`, `RawSettings` and `ElmToolchain` all fit together, but  at least now the user will have more control.

Tomorrow I plan on making it easier for the user to fix problems like trying to run `elm-test` via Run Configuration when the path to `elm-test` is unknown. I will also standardize the notification UI for these sorts of things.

### DIFF
--- a/src/main/java/org/frawa/elmtest/run/ElmTestRunProfileState.java
+++ b/src/main/java/org/frawa/elmtest/run/ElmTestRunProfileState.java
@@ -3,9 +3,7 @@ package org.frawa.elmtest.run;
 import com.intellij.execution.ExecutionException;
 import com.intellij.execution.Executor;
 import com.intellij.execution.configurations.CommandLineState;
-import com.intellij.execution.configurations.GeneralCommandLine;
 import com.intellij.execution.configurations.RunConfiguration;
-import com.intellij.execution.process.ColoredProcessHandler;
 import com.intellij.execution.process.ProcessHandler;
 import com.intellij.execution.runners.ExecutionEnvironment;
 import com.intellij.execution.testframework.TestConsoleProperties;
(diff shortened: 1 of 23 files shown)


klazuka:
FYI @frawa and @FloWi

FloWi:
Looks great! I really like the auto-discover feature! Works like a charm. Is it common to use the compiler in the node_modules first? I didn't know elm gets also installed there and was surprised to discover it.

I noticed a small annoyance - when you first open the toolchain-settings the whole dialog is a bit too small for its content. For some reason I cannot reproduce it now, since IDEA somewhere stores the size of the dialog after manually resizing it and I haven't found a way to reset that.

klazuka:
The tools are only installed in a `node_modules` directory if you ran something like `npm install --save elm` from the command line (note the `--save` part). Typically people do this so that their build server will also use the same tools.

As for the size of the settings window, yeah, I know, I haven't been able to figure out how to tell the settings dialog to resize to fit its content so that there's no scrolling. I'm not even sure that it's possible.

FloWi:
> As for the size of the settings window, yeah, I know, I haven't been able to figure out how to tell the settings dialog to resize to fit its content so that there's no scrolling. I'm not even sure that it's possible.
Same here. Also haven't found to figure out how to do that.

PR #328, opened 2019-02-27
Fix/324 elm test elm make

frawa:
and support producing test configs from nested test projects


frawa:
This is quick fixes to some issues I observed with 0.18 projects.
Thinking about ways to add tests now ...

klazuka:
@frawa testing Elm 0.18 is hard because of the way that `elm-stuff` is setup, especially with the tests project nested within the container project. I'm ok with just manually testing this by hand. Hopefully sometime this year we can remove Elm 0.18 support entirely.

But if you figured out a relatively easy way to add an integration test, that would be good. Just let me know either way.

frawa:
Better.

Issue #324, opened 2019-02-27
elm-test 0.18 needs to use elm-make


frawa:
Elm tests are launched using (depending on config)
```
/usr/local/bin/elm-test --report=json --compiler /usr/local/bin/elm
```

With 0.18, I observe a situtation, where it needs to be called
```
/usr/local/bin/elm-test --report=json --compiler /usr/local/bin/elm-make
```

klazuka:
This was fixed by #328

PR #332, opened 2019-02-28
WIP: use Elm compiler to view errors

### LINKED ISSUE #83: Retrieve compiler output color information
When compiling in webstorm ( via hotkeys ), the console's output is in one color (even though it supports colors). In elm 0.19, some error messages are only useful in color:
https://github.com/elm/error-message-catalog/issues/269

Evan mentioned that the compiler colors can be found in the following package:
https://package.elm-lang.org/packages/elm/project-metadata-utils/latest/

It would be _awesome_ if this plugin could relay this information to the console.

More info:
Webstorm allows you to configure colors and font for their console. 
There are plugins for embedding a terminal, even a iTerm.
But you cannot link the output of a external tool to a terminal, and you cannot set the console to be colored like a terminal.

klazuka:
Creating this PR for @clojj so that I can provide feedback directly on the code.

Fixes #83


klazuka:
I loaded [elm-spa-example](https://github.com/rtfeldman/elm-spa-example), `src/Page/Article/Editor.elm` and commented out the `Model` type declaration. Then I built the project and clicked around in the list of errors. I got this exception:
```
kotlin.KotlinNullPointerException
	at org.elm.ide.toolwindow.ElmCompilerPanel.getData(ElmCompilerPanel.kt:161)
	at com.intellij.ide.impl.DataManagerImpl.getDataFromProvider(DataManagerImpl.java:72)
	at com.intellij.ide.impl.DataManagerImpl.lambda$getDataFromProvider$0(DataManagerImpl.java:79)
	at com.intellij.openapi.actionSystem.DataKey.getData(DataKey.java:56)
	at com.intellij.ide.impl.dataRules.NavigatableArrayRule.getData(NavigatableArrayRule.java:12)
	at com.intellij.ide.impl.DataManagerImpl.getDataFromProvider(DataManagerImpl.java:79)
	at com.intellij.ide.impl.DataManagerImpl.getData(DataManagerImpl.java:59)
	at com.intellij.ide.impl.DataManagerImpl.access$200(DataManagerImpl.java:44)
	at com.intellij.ide.impl.DataManagerImpl$MyDataContext.doGetData(DataManagerImpl.java:371)
	at com.intellij.ide.impl.DataManagerImpl$MyDataContext.getData(DataManagerImpl.java:348)
	at com.intellij.openapi.actionSystem.DataKey.getData(DataKey.java:50)
	at com.intellij.ui.AutoScrollToSourceHandler.lambda$null$2(AutoScrollToSourceHandler.java:168)
	at com.intellij.openapi.application.TransactionGuardImpl.runSyncTransaction(TransactionGuardImpl.java:88)
	at com.intellij.openapi.application.TransactionGuardImpl.lambda$submitTransaction$1(TransactionGuardImpl.java:111)
	at com.intellij.openapi.application.TransactionGuardImpl.submitTransaction(TransactionGuardImpl.java:120)
	at com.intellij.openapi.application.TransactionGuard.submitTransaction(TransactionGuard.java:122)
	at com.intellij.ui.AutoScrollToSourceHandler.lambda$scrollToSource$3(AutoScrollToSourceHandler.java:157)
	at com.intellij.openapi.util.ExecutionCallback.doWhenExecuted(ExecutionCallback.java:100)
	at com.intellij.openapi.util.ActionCallback.doWhenDone(ActionCallback.java:111)
	at com.intellij.ui.AutoScrollToSourceHandler.scrollToSource(AutoScrollToSourceHandler.java:157)
	at com.intellij.ui.AutoScrollToSourceHandler.lambda$onMouseClicked$0(AutoScrollToSourceHandler.java:128)
	at com.intellij.openapi.application.TransactionGuardImpl$2.run(TransactionGuardImpl.java:315)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.doRun(LaterInvocator.java:435)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.runNextEvent(LaterInvocator.java:419)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.run(LaterInvocator.java:403)
        ... SNIP ...
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:93)
	at java.awt.EventDispatchThread.run(EventDispatchThread.java:82)
```

klazuka:
It's hard to read the gray text in this error output:
screen shot 2019-02-28 at 3 47 46 pm

Is it possible to control the background and foreground colors to increase the contrast?

klazuka:
@clojj thanks for putting this together! 👍 

Here are some things that I noticed while trying it out:

- [ ] the forward and backward buttons should be enabled/disabled contextually based on whether there is a previous/next error
- [ ] the standard IntelliJ keyboard shortcuts for moving between items in a toolwindow list should work for moving between errors. For instance, in the "Find Usages" panel on my Mac, `Cmd-Option-DownArrow` moves to the next thing in the list.
- [ ] when I select "Build Elm Compiler" from IntelliJ's "Tools" menu, nothing happens. I would expect the tool window to popup from the bottom
- [ ] if the path to the Elm compiler is not set, the error message should include a button/hyperlink to open the Elm toolchain settings so that the user can fix the problem. Probably should use a balloon notification like `ElmExternalFormatAction` instead of a dialog.
- [ ] the text in the list of errors does not need to be red: just make it black
- [ ] the error list should be sorted by module name, then by line number and then by column
- [ ] selecting an error should also navigate to that line/column in the editor ("autoscroll to source"). We may want to make this configurable, but I think that—for now—we should just have it be always on.

klazuka:
screen shot 2019-02-28 at 3 52 53 pm

The "Elm Compiler" group should be nested within the "Elm" group. I think you can control this in the `plugin.xml`.

Also, the back and forward actions should have text.

src/main/kotlin/org/elm/utils/CircularList.kt
```diff
@@ -0,0 +1,51 @@
...
+    private val maxIndex = list.size - 1
+
+    private var index: Int = 0
+
+    fun get(): T = list[index]
+
+    fun getIndex(): Int = index
+
+    fun set(i: Int) {
+        index = i
```
klazuka: I think this needs a bounds check.

Also might want to rename the function to something like `jumpTo` as, to me, "set" implies that we are changing a value in the collection.

src/main/kotlin/org/elm/utils/CircularList.kt
```diff
@@ -0,0 +1,51 @@
...
+        safelyAdd(1)
+        return get()
+    }
+
+    fun prev(): T {
+        safelyAdd(-1)
+        return get()
+    }
+
+    private fun safelyAdd(n: Int) {
```
klazuka: rename to `advance`?

src/main/kotlin/org/elm/workspace/compiler/ElmBackAction.kt
```diff
@@ -0,0 +1,29 @@
+package org.elm.workspace.compiler
+
+import com.intellij.icons.AllIcons
+import com.intellij.openapi.actionSystem.AnAction
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.util.messages.Topic
+import org.elm.ide.toolwindow.ElmCompilerPanel
+
+class ElmBackAction(private val elmCompilerPanel: ElmCompilerPanel) : AnAction() {
```
klazuka: I'm no expert on the IntelliJ action system, but I'm pretty sure that they are supposed to be "beans" (no-arg constructor).

I took a brief look at the implementors of the `AnAction` abstract class, and, as far as I can tell, everything uses the  `DataContext` system for acquiring information from other parts of the system.

Last fall I made some notes on how `DataContext` and company work. Here's a gist: https://gist.github.com/klazuka/23a9894253fb6878f95ab74ba1b75bd9

So rather than taking the `ElmCompilerPanel` in the constructor, I think you should instead be obtaining it from the `DataContext` (which also probably means that on the `ElmCompilerPanel` side you will need to somehow "register" it).

src/main/kotlin/org/elm/workspace/compiler/ElmBackAction.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    override fun update(e: AnActionEvent) {
+        e.presentation.icon = AllIcons.Actions.Back
+        e.presentation.isEnabled = !elmCompilerPanel.compilerMessages.isEmpty()
+    }
+
+    override fun actionPerformed(e: AnActionEvent) {
+        val project = e.project
+                ?: return
+        project.messageBus.syncPublisher(ERRORS_BACK_TOPIC).back()
```
klazuka: I guess this is OK, but you may want to check how other JetBrains actions are implemented to see how they do it.

I would bet that there are a bunch of examples in the codebase where they handle "back/forward"-style actions.

src/main/kotlin/org/elm/workspace/commandLineTools/ElmCLI.kt
```diff
@@ -17,13 +17,11 @@ import java.nio.file.Path
  */
 class ElmCLI(private val elmExecutablePath: Path) {
 
-    // TODO [kl] allow the caller to specify the main entry point Elm file
-
-    fun make(owner: Disposable, elmProject: ElmProject): ProcessOutput {
+    fun make(owner: Disposable, elmProject: ElmProject, path: String): ProcessOutput {
```
klazuka: please use Java's `Path` type instead of a string for consistency with the rest of the plugin.

src/main/kotlin/org/elm/workspace/compiler/ElmBuildAction.kt
```diff
@@ -0,0 +1,81 @@
...
+                ?: return
+        saveAllDocuments()
+
+        val elmCLI = project.elmToolchain.elmCLI
+        if (elmCLI == null) {
+            Messages.showErrorDialog("No path to the Elm compiler", "Build Error")
+            return
+        }
+
+        val elmProject = project.elmWorkspace.allProjects.first()
```
klazuka: Please add a TODO comment to handle the multi-project case. 

At a minimum we should use the `DataContext` to get the current editor/document/psifile, and then use `org.elm.workspace.ElmWorkspaceService#findProjectForFile` to determine the "active" Elm project.

But it might be better to just show a picker/popover UI listing all the Elm projects and letting the user pick.

src/main/kotlin/org/elm/workspace/compiler/ElmBuildAction.kt
```diff
@@ -0,0 +1,81 @@
...
+
+        // find "main" function
+        val clientLocation = LookupClientLocation(project, elmProject)
+        val elements = ElmLookup.findByName<ElmNamedElement>("main", clientLocation)
```
klazuka: rather than restricting the return type to `ElmNamedElement`, instead use `ElmFunctionDeclarationLeft`

src/main/kotlin/org/elm/workspace/compiler/ElmBuildAction.kt
```diff
@@ -0,0 +1,81 @@
...
+        val mainElements = elements
+                .filter { element -> element.findInference() != null && element.findInference()?.ty is TyUnion }
+                .filter { element -> elmMainTypes.contains(Pair((element.findInference()?.ty as TyUnion).module, (element.findInference()?.ty as TyUnion).name)) }
+
+        if (mainElements.isNotEmpty()) {
```
klazuka: Please flip the conditional so that the error case is handled first, showing the dialog and returning early. Then you can remove the nesting from the happy path.

klazuka:
Testing IntelliJ stuff is not well documented. But the good news is that [intellij-rust](https://github.com/intellij-rust/intellij-rust) is a gold mind of solutions. They have a *lot* of tests, so I would definitely see if they have any features which are comparable to this and then see how they tested it. Since this involves an external dependency, you should also look at how I'm handling it for `elm-format`. See `ElmExternalFormatActionTest`.

build.gradle
```diff
@@ -141,8 +142,10 @@ dependencies {
         exclude module: "kotlin-runtime"
         exclude module: "kotlin-stdlib"
     }
+    compile "com.beust:klaxon:5.0.1"
```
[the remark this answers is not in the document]
clojj: yep.. gson it will be then

src/main/kotlin/org/elm/utils/CircularList.kt
clojj: Also it could make more sense, not to sift circularly but stoppen at each end... ?

src/main/kotlin/org/elm/workspace/compiler/ElmBackAction.kt
clojj: Felt weird doing this anyway 😉

clojj:
Some color mapping may be necessary

it is not a priority, but I am considering using the editors font for the message pane on the right.. this would make the codesnippets more consistent ( e.g. ligatures!)

catz:
Hello,

Can you please describe how to setup project compilation ? I see only Elm Test in Edit Configurations and it runs em-test.
I mean I want two targets like debug: elm make src/Main.elm --output elm.js --debug
and release: elm make src/Main.elm --output elm.js

I'm new to idea and maybe there is simple instruction about it? Thank you.

src/main/kotlin/org/elm/utils/CircularList.kt
klazuka: yes, then we could get rid of this entirely. And I think it would make more sense with the list view that you have.

klazuka:
@catz in the currently released version of the plugin, the only "run configuration" that we support is running tests. This PR will add support for running `elm make` in order to check errors, but not necessarily to generate code or create a runnable webapp. However, in the future I intend to add a way to compile and run your Elm app, with optional hot module reloading.

catz:
@klazuka got it. added run external tool into elm-test configuration, so far so good.

clojj:
> I loaded [elm-spa-example](https://github.com/rtfeldman/elm-spa-example), `src/Page/Article/Editor.elm` and commented out the `Model` type declaration. Then I built the project and clicked around in the list of errors. I got this exception:
> 
> ```
> kotlin.KotlinNullPointerException
> 	at org.elm.ide.toolwindow.ElmCompilerPanel.getData(ElmCompilerPanel.kt:161)
> 	at com.intellij.ide.impl.DataManagerImpl.getDataFromProvider(DataManagerImpl.java:72)


following your steps, I got this

```
com.intellij.psi.PsiInvalidElementAccessException: Element: class org.elm.lang.core.psi.elements.ElmTypeAliasDeclaration #Elm  because: parent is null
invalidated at: see attachment
	at com.intellij.psi.util.PsiUtilCore.ensureValid(PsiUtilCore.java:482)
	at com.intellij.psi.impl.source.resolve.ResolveCache.resolve(ResolveCache.java:158)
	at com.intellij.psi.impl.source.resolve.ResolveCache.resolveWithCaching(ResolveCache.java:239)
	at org.elm.lang.core.resolve.reference.ElmReferenceCached.resolve(ElmReferenceCached.kt:15)
	at org.elm.lang.core.resolve.scope.ImportScope.getExposedTypes(ImportScope.kt:92)
	at org.elm.lang.core.resolve.scope.ModuleScope.getVisibleImportTypes(ModuleScope.kt:185)
	at org.elm.lang.core.resolve.scope.ModuleScope.access$getVisibleImportTypes(ModuleScope.kt:45)
	at org.elm.lang.core.resolve.scope.ModuleScope$getVisibleTypes$1.compute(ModuleScope.kt:176)
	at com.intellij.psi.util.CachedValuesManager.lambda$getCachedValue$0(CachedValuesManager.java:140)
	at com.intellij.psi.impl.PsiCachedValueImpl.doCompute(PsiCachedValueImpl.java:49)
	at com.intellij.util.CachedValueBase.getValueWithLock(CachedValueBase.java:222)
	at com.intellij.psi.impl.PsiCachedValue.getValueWithLock(PsiCachedValue.java:60)
	at com.intellij.psi.impl.PsiCachedValueImpl.getValue(PsiCachedValueImpl.java:38)
	at com.intellij.psi.util.CachedValuesManager.getCachedValue(CachedValuesManager.java:136)
	at org.elm.lang.core.resolve.scope.ModuleScope.getVisibleTypes(ModuleScope.kt:172)
	at org.elm.lang.core.resolve.reference.SimpleTypeReference.getCandidates(SimpleTypeReference.kt:20)
	at org.elm.lang.core.resolve.reference.SimpleTypeReference.resolveInner(SimpleTypeReference.kt:17)
	at org.elm.lang.core.resolve.reference.ElmReferenceCached$Resolver.resolve(ElmReferenceCached.kt:20)
	at org.elm.lang.core.resolve.reference.ElmReferenceCached$Resolver.resolve(ElmReferenceCached.kt:18)
	at com.intellij.psi.impl.source.resolve.ResolveCache.lambda$resolve$0(ResolveCache.java:150)
	at com.intellij.openapi.util.RecursionManager$2.doPreventingRecursion(RecursionManager.java:98)
	at com.intellij.psi.impl.source.resolve.ResolveCache.resolve(ResolveCache.java:149)
	at com.intellij.psi.impl.source.resolve.ResolveCache.resolveWithCaching(ResolveCache.java:239)
	at org.elm.lang.core.resolve.reference.ElmReferenceCached.resolve(ElmReferenceCached.kt:15)
	at org.elm.lang.core.types.TypeExpression.typeRefType(TypeExpressionInference.kt:198)
	at org.elm.lang.core.types.TypeExpression.typeSignatureDeclType(TypeExpressionInference.kt:183)
	at org.elm.lang.core.types.TypeExpression.typeRefType(TypeExpressionInference.kt:197)
	at org.elm.lang.core.types.TypeExpression.typeSignatureDeclType(TypeExpressionInference.kt:183)
	at org.elm.lang.core.types.TypeExpression.access$typeSignatureDeclType(TypeExpressionInference.kt:99)
	at org.elm.lang.core.types.TypeExpression$typeExpresionType$segments$1.invoke(TypeExpressionInference.kt:168)
	at org.elm.lang.core.types.TypeExpression$typeExpresionType$segments$1.invoke(TypeExpressionInference.kt:99)
	at kotlin.sequences.TransformingSequence$iterator$1.next(Sequences.kt:149)
	at kotlin.sequences.SequencesKt___SequencesKt.toCollection(_Sequences.kt:647)
	at kotlin.sequences.SequencesKt___SequencesKt.toMutableList(_Sequences.kt:677)
	at kotlin.sequences.SequencesKt___SequencesKt.toList(_Sequences.kt:668)
	at org.elm.lang.core.types.TypeExpression.typeExpresionType(TypeExpressionInference.kt:169)
	at org.elm.lang.core.types.TypeExpression.beginTypeRefInference(TypeExpressionInference.kt:110)
	at org.elm.lang.core.types.TypeExpressionInferenceKt$typeExpressionInference$cachedValue$5.compute(TypeExpressionInference.kt:65)
	at com.intellij.psi.util.CachedValuesManager.lambda$getCachedValue$0(CachedValuesManager.java:140)
	at com.intellij.psi.impl.PsiCachedValueImpl.doCompute(PsiCachedValueImpl.java:49)
	at com.intellij.util.CachedValueBase.getValueWithLock(CachedValueBase.java:222)
	at com.intellij.psi.impl.PsiCachedValue.getValueWithLock(PsiCachedValue.java:60)
	at com.intellij.psi.impl.PsiCachedValueImpl.getValue(PsiCachedValueImpl.java:38)
	at com.intellij.psi.util.CachedValuesManager.getCachedValue(CachedValuesManager.java:136)
	at org.elm.lang.core.types.TypeExpressionInferenceKt.typeExpressionInference(TypeExpressionInference.kt:64)
	at org.elm.lang.core.types.InferenceScope.bindFunctionDeclarationParameters(TypeInference.kt:650)
	at org.elm.lang.core.types.InferenceScope.bindParameters(TypeInference.kt:630)
	at org.elm.lang.core.types.InferenceScope.beginDeclarationInference(TypeInference.kt:103)
	at org.elm.lang.core.types.TypeInferenceKt$inference$1.compute(TypeInference.kt:36)
	at com.intellij.psi.util.CachedValuesManager.lambda$getCachedValue$0(CachedValuesManager.java:140)
	at com.intellij.psi.impl.PsiCachedValueImpl.doCompute(PsiCachedValueImpl.java:49)
	at com.intellij.util.CachedValueBase.getValueWithLock(CachedValueBase.java:222)
	at com.intellij.psi.impl.PsiCachedValue.getValueWithLock(PsiCachedValue.java:60)
	at com.intellij.psi.impl.PsiCachedValueImpl.getValue(PsiCachedValueImpl.java:38)
	at com.intellij.psi.util.CachedValuesManager.getCachedValue(CachedValuesManager.java:136)
	at org.elm.lang.core.types.TypeInferenceKt.inference(TypeInference.kt:32)
	at org.elm.lang.core.types.TypeInferenceKt.findInference(TypeInference.kt:24)
	at org.elm.workspace.compiler.ElmBuildAction.actionPerformed(ElmBuildAction.kt:41)
	at com.intellij.openapi.actionSystem.ex.ActionUtil$1.run(ActionUtil.java:258)
	at com.intellij.openapi.actionSystem.ex.ActionUtil.performActionDumbAware(ActionUtil.java:275)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher$1.performAction(IdeKeyEventDispatcher.java:618)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher.lambda$processAction$2(IdeKeyEventDispatcher.java:667)
	at com.intellij.openapi.application.TransactionGuardImpl.performUserActivity(TransactionGuardImpl.java:195)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher.processAction(IdeKeyEventDispatcher.java:666)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher.processActionOrWaitSecondStroke(IdeKeyEventDispatcher.java:519)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher.inInitState(IdeKeyEventDispatcher.java:474)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher.dispatchKeyEvent(IdeKeyEventDispatcher.java:211)
	at com.intellij.ide.IdeEventQueue._dispatchEvent(IdeEventQueue.java:643)
	at com.intellij.ide.IdeEventQueue.dispatchEvent(IdeEventQueue.java:363)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:201)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:105)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:101)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:93)
	at java.awt.EventDispatchThread.run(EventDispatchThread.java:82)
```

klazuka:
I was also seeing that exception, but I assumed that it was (somehow) being caused indirectly by the null pointer exception. Let me know if it still occurs after the other bug is fixed and then I will dig in to make sure that we're following the threading rules, etc.

clojj:
ok, think I found it.... it has to be like this
`                    val file = LocalFileSystem.getInstance().findFileByPath(Paths.get(project.basePath + "/" + compilerMessages.get().path))
`

 maybe there's a simpler way to get the VirtualFile, I will see

build.gradle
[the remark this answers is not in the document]
clojj: gson parsing is going to work nicely, I just tested...

clojj:
For this one I could not yet find any examples of how to do it.... ?

> the standard IntelliJ keyboard shortcuts for moving between items in a toolwindow list should work for moving between errors. For instance, in the "Find Usages" panel on my Mac, Cmd-Option-DownArrow moves to the next thing in the list.

clojj:
I will continue with the topic plumbing and refactoring...

klazuka:
I will take a look around to see if I can figure out how JetBrains does the keybindings for moving previous/next in the tool window.

clojj:
If we can leverage IJ‘s Next/Prev System, I‘d be happy to remove these self-built Back/Forward actions altogether, yes

klazuka:
Ok, here's how it works.

When building your toolbar:
- use `com.intellij.ide.impl.CommonActionsManagerImpl#createNextOccurenceAction` to construct the button that will move to the next error
- that function takes an `OccurrenceNavigator`. You can either implement the interface directly or you can use `OccurenceNavigatorSupport`. However, as best I can tell, the latter is overkill for this situation.

For an example, see `com.intellij.ide.todo.TodoPanel#initUI`

Also, I learned something new. JetBrains does have actions which take an argument. Look at the constructor for `com.intellij.ide.actions.NextOccurenceToolbarAction` where it then uses `copyFrom` to base the action off an existing (presumably no-arg) action.

clojj:
@klazuka 
works as advertised !
let me clean things up a bit.. will push soon

clojj:
works for me now... the only part I don't like at the moment is this one
https://github.com/clojj/intellij-elm/blob/d0e252f4e0297f43c35ee58d1d10614e3b6ae9d2/src/main/kotlin/org/elm/ide/toolwindow/ElmCompilerPanel.kt#L41

src/main/kotlin/org/elm/ide/toolwindow/ElmCompilerPanel.kt
```diff
@@ -0,0 +1,211 @@
...
+import javax.swing.JList
+import javax.swing.JTextPane
+import javax.swing.ListSelectionModel
+
+
+class ElmCompilerPanel(private val project: Project, private val contentManager: ContentManager) : SimpleToolWindowPanel(true, false), Disposable, OccurenceNavigator {
+
+    private fun occurenceInfo(): OccurenceNavigator.OccurenceInfo {
+        var filePath = compilerMessages[indexCompilerMessages].path
+        if (!filePath.startsWith("/")) filePath = project.basePath + "/" + filePath // TODO any simpler way for this ?
```
klazuka: I have `org.elm.openapiext.UtilsKt#findFileByMaybeRelativePath` which I copied from intellij-rust. That might be kinda helpful here. Or maybe create another variant where the receiver is an IntelliJ `Project`?

As for the dance when moving from VirtualFile to Psi and Document, I think that's just the way it is. However, I do have one utility function, `org.elm.openapiext.UtilsKt#toPsiFile` which should be helpful  (that also came from intellij-rust).

src/main/kotlin/org/elm/ide/toolwindow/ElmCompilerPanel.kt
```diff
@@ -0,0 +1,211 @@
...
+        val file = LocalFileSystem.getInstance().findFileByPath(Paths.get(filePath))
+        val psiFile = PsiManager.getInstance(project).findFile(file!!)
```
klazuka: it would be nice if we avoided `!!`. Does the caller expect a non-null return value? If not, just return null early.

src/main/kotlin/org/elm/ide/toolwindow/ElmCompilerPanel.kt
```diff
@@ -0,0 +1,211 @@
...
+        toolbar.setTargetComponent(this)
+        return toolbar.component
+    }
+
+    override fun getData(dataId: String): Any? {
+        return when {
+            CommonDataKeys.NAVIGATABLE.`is`(dataId) -> {
+                if (!compilerMessages.isEmpty()) {
+                    var filePath = compilerMessages[indexCompilerMessages].path
+                    if (!filePath.startsWith("/")) filePath = project.basePath + "/" + filePath
```
klazuka: this code appears to duplicate a lot of the logic around line 43 above.

clojj:
still some issue with `!!`, but it looks much better now

klazuka:
Thanks. I will checkout your branch and try it out in the next couple of days. I don't have much freetime right now, and I need to prioritize getting the 3.0.1 bugfix release out.

clojj:
no problem.. have some todos and ideas to work on

clojj:
I also think that focus should now be on the user experience... for instance I wonder what information should really go into that list on the left panel, I think line+column is not really that useful because of the automatic navigation

clojj:
@klazuka 
Hi, just pushed something that finds the file everytime (so far, :) )
to get rid of `!!` I did this (in case there is null, which is never expected):
`throw RuntimeException("The impossible happened...")`

klazuka:
I just tried it out now. Overall, it's great! It runs super fast and the keyboard navigation works well. 🔥 

I'm still seeing the `PsiInvalidElementAccessException` when commenting out type declarations and function declarations. This is probably something that I'm going to have to dig into, as it might be a deeper problem in the reference system. Dunno yet. I've never had a problem with this before.

Things that could use a little polish:
- sort errors by line number within a file
- if there are no errors, then don't show the white&black split. right now it's too intense on the eyese (ultimately I think we'd want to only use the dark-background error display only when the user is running a "night mode" color scheme, but that is probably a bigger job, and it can be done later)
- check to see if there's a way to hook into IntelliJ's "Build Project" action (that way the user doesn't need to learn yet-another keybinding)
- when user tries to invoke `elm make` but the path to the compiler is missing/bad, the error dialog should include a hyperlink that opens the Elm settings UI (see `ElmExternalFormatAction`)

You're almost there! Thanks for sticking with it 💯

clojj:
about PsiInvalidElementAccessException

...it could be these "refresh" options I see all over the file related APIs.... maybe

I will try to reproduce this as I did not run into this recently

clojj:
hook into "Ctrl F9" could be this maybe:
´´´
<action id="Rust.Build"
                class="org.rust.ide.actions.RsBuildAction"
                text="Build"
                description="Build Project"
                icon="AllIcons.Actions.Compile"
                use-shortcut-of="CompileDirty">
            <add-to-group group-id="ToolbarRunGroup" relative-to-action="RunConfiguration" anchor="before"/>
            <add-to-group group-id="TouchBarDefault" relative-to-action="RunConfiguration" anchor="before"/>
</action>
´´´

klazuka:
> it could be these "refresh" options I see all over the file related APIs

Quite possibly. That part I do not understand well, but intellij-rust seems to do it often in these "action" handlers.

As for the `use-shortcut-of="CompileDirty"`, yeah, that looks promising. I wonder how that works when you have multi-language projects? For instance, I have an IntelliJ project which consists of a Kotlin backend and an Elm frontend. When I click "Build Project" is it going to run both actions? Just one? Or maybe the trick is that you have to set `presentation.enabled` based on context so that if the user is editing an Elm file it triggers the Elm build, and if you're editing a Kotlin file it triggers a Kotlin build?

clojj:
@klazuka 
I can reproduce
`com.intellij.psi.PsiInvalidElementAccessException: Element: class org.elm.lang.core.psi.elements.ElmTypeAliasDeclaration #Elm  because: parent is null
`
as follows:
- open elm-spa-example (version without errors)
- open file `src/Page/Article/Editor.elm`
- comment out `type alias Model = ...` (using block-comment)
... immediately there is this Exception, I did not yet do 'Build Elm"

clojj:
I added your suggested improvements..
still open:
- issue with `com.intellij.psi.PsiInvalidElementAccessException`
- invoke by regular "Build" / "Ctrl-F9" (that plugin.xml change did not work)

I think your point about multi-language/multi-build issues are valid, I think this has to be investigated some more. That annoying exception should have priority, I think

clojj:
ah.. the sorting doesn't work yet... I'm on it

clojj:
sorting is fixed now

klazuka:
You're right! The `PsiInvalidElementAccessException` is unrelated. I created a separate ticket for it (#344), and I will look into it.

So I think what remains for this PR is:
- hooking into the standard "Build" keyboard shortcut (if possible)
- test with Elm 18 (we ran into problems with the elm-test runner, see #324)

klazuka:
The changes to the UI look good. I like that it doesn't have the white & black contrast problems anymore. And the "zero errors" case looks a lot better.

Two more things:
- can you add a try/catch around the call to `ElmCLI#make` and handle `com.intellij.execution.ExecutionException`? Right now if you type in a bogus path for the Elm compiler in the toolchain settings, an uncaught exception will be thrown all the way up to the top. It would be better if we handled it and display a friendly notification with a hyperlink to open the toolchain settings to fix the path
- can you add some padding/inset to the right-side panel? ideally the padding would match on both sides of the split

screen shot 2019-03-06 at 6 17 27 pm

clojj:
@klazuka 
insets + elm make exception fixed

clojj:
I changed the JBList to JBTable for a more pleasant alignment:
error-table

clojj:
error-table

klazuka:
@clojj the new UI looks great!

I did run into an exception when trying to navigate between errors using `Cmd-Option-DownArrow`:
```
java.lang.RuntimeException: The impossible happened... virtualFile is null for '../src/JsonTree.elm'
	at org.elm.ide.toolwindow.ElmCompilerPanel.startFromErrorMessage(ElmCompilerPanel.kt:262)
	at org.elm.ide.toolwindow.ElmCompilerPanel.updateSelectionAndCreateOccurenceInfo(ElmCompilerPanel.kt:58)
	at org.elm.ide.toolwindow.ElmCompilerPanel.goNextOccurence(ElmCompilerPanel.kt:79)
	at com.intellij.ide.actions.NextOccurenceAction.go(NextOccurenceAction.java:15)
	at com.intellij.ide.actions.OccurenceNavigatorActionBase.actionPerformed(OccurenceNavigatorActionBase.java:38)
	at com.intellij.ide.actions.NextOccurenceAction.actionPerformed(NextOccurenceAction.java:7)
	at com.intellij.openapi.actionSystem.ex.ActionUtil$1.run(ActionUtil.java:258)
	at com.intellij.openapi.actionSystem.ex.ActionUtil.performActionDumbAware(ActionUtil.java:275)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher$1.performAction(IdeKeyEventDispatcher.java:618)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher.lambda$processAction$2(IdeKeyEventDispatcher.java:667)
	at com.intellij.openapi.application.TransactionGuardImpl.performUserActivity(TransactionGuardImpl.java:195)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher.processAction(IdeKeyEventDispatcher.java:666)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher.processActionOrWaitSecondStroke(IdeKeyEventDispatcher.java:519)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher.inInitState(IdeKeyEventDispatcher.java:474)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher.dispatchKeyEvent(IdeKeyEventDispatcher.java:211)
	at com.intellij.ide.IdeEventQueue._dispatchEvent(IdeEventQueue.java:643)
	at com.intellij.ide.IdeEventQueue.dispatchEvent(IdeEventQueue.java:363)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:201)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:105)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:101)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:93)
	at java.awt.EventDispatchThread.run(EventDispatchThread.java:82)
```

Maybe it doesn't like that relative path (`../src/JsonTree.elm`)? This was from my [elm-json-tree-view](https://github.com/Microsoft/elm-json-tree-view) package.

klazuka:
Here's another exception I ran into:
```
java.lang.IndexOutOfBoundsException: Empty list doesn't contain element at index 0.
	at kotlin.collections.EmptyList.get(Collections.kt:36)
	at kotlin.collections.EmptyList.get(Collections.kt:24)
	at org.elm.ide.toolwindow.ElmCompilerPanel$$special$$inlined$apply$lambda$1.valueChanged(ElmCompilerPanel.kt:132)
	at javax.swing.DefaultListSelectionModel.fireValueChanged(DefaultListSelectionModel.java:184)
	at javax.swing.DefaultListSelectionModel.fireValueChanged(DefaultListSelectionModel.java:164)
	at javax.swing.DefaultListSelectionModel.fireValueChanged(DefaultListSelectionModel.java:211)
	at javax.swing.DefaultListSelectionModel.changeSelection(DefaultListSelectionModel.java:405)
	at javax.swing.DefaultListSelectionModel.changeSelection(DefaultListSelectionModel.java:415)
	at javax.swing.DefaultListSelectionModel.setSelectionInterval(DefaultListSelectionModel.java:459)
	at javax.swing.JTable.setRowSelectionInterval(JTable.java:2166)
	at org.elm.ide.toolwindow.ElmCompilerPanel$$special$$inlined$with$lambda$1$1.run(ElmCompilerPanel.kt:200)
	at com.intellij.openapi.application.TransactionGuardImpl$2.run(TransactionGuardImpl.java:315)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.doRun(LaterInvocator.java:435)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.runNextEvent(LaterInvocator.java:419)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.run(LaterInvocator.java:403)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:762)
	at java.awt.EventQueue.access$500(EventQueue.java:98)
	at java.awt.EventQueue$3.run(EventQueue.java:715)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:80)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:732)
	at com.intellij.ide.IdeEventQueue.defaultDispatchEvent(IdeEventQueue.java:719)
	at com.intellij.ide.IdeEventQueue._dispatchEvent(IdeEventQueue.java:668)
	at com.intellij.ide.IdeEventQueue.dispatchEvent(IdeEventQueue.java:363)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:201)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:105)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:101)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:93)
	at java.awt.EventDispatchThread.run(EventDispatchThread.java:82)
```

This happened after trying to invoke the Elm build from the elm-spa-example project back when it was still an Elm 0.18 project (sha 298bfdf). Not sure if the bug is related to Elm 0.18 or not.

We don't strictly *have* to support Elm 0.18 with this feature (especially if it would be a lot work if, say, the JSON report output is different), but we should at least show an error message that explains that it's not supported.

clojj:
can you retry ? I think I fixed it

clojj:
@klazuka 
to also support 'elm-make' from Elm 0.18, we should move this fun
`org.elm.workspace.commandLineTools.ElmFormatCLI.Companion#getElmVersion`
...so I can reuse it in `org.elm.workspace.commandLineTools.ElmCLI#make`

clojj:
@klazuka 
just ran into that 1st exception... I will take a detailed look at it

build.gradle
```diff
@@ -18,6 +18,7 @@ plugins {
 }
 
 repositories {
+    jcenter()
```
klazuka: this can probably be removed (presumably it came from the klaxon dependency?)

src/main/kotlin/org/elm/ide/toolwindow/ElmCompilerPanel.kt
```diff
@@ -0,0 +1,298 @@
...
+import javax.swing.ScrollPaneConstants.HORIZONTAL_SCROLLBAR_NEVER
+import javax.swing.ScrollPaneConstants.VERTICAL_SCROLLBAR_AS_NEEDED
+import javax.swing.border.EmptyBorder
+import javax.swing.table.DefaultTableCellRenderer
+import javax.swing.table.DefaultTableModel
+
+
+class ElmCompilerPanel(private val project: Project, private val contentManager: ContentManager) : SimpleToolWindowPanel(true, false), Disposable, OccurenceNavigator {
+
+    // TODO: support multiple projects ? how is 'elm make' supported for multiple projects anyway ?
```
klazuka: The ElmProject should be determined based on the file currently open in the IntelliJ editor. Use IntelliJ's DataManager / DataContext system to find it (not sure exactly how this will work in the case of a split editor). Then use `org.elm.workspace.ElmWorkspaceService#findProjectForFile` to lookup the `ElmProject` for the current `.elm` file.

Note: there may be no active Elm file, and there may be no attached ElmProjects. In both cases, we should just display an error.

src/main/kotlin/org/elm/ide/toolwindow/ElmCompilerPanel.kt
```diff
@@ -0,0 +1,298 @@
...
+    lateinit var elmProjects: List<ElmProject>
```
klazuka: `lateinit` is a bad idea here. I would rather that we just initialize it to an empty list. But even the idea of caching the list of projects here is itself not a good idea. The truth must live in the `ElmWorkspaceService`. What happens if a user detaches a project? Or adds a new one? We don't want things to get out-of-sync.

clojj:
I think that relative paths like `../src/SomeThing.elm` are fixed now

clojj:
The json from 0.18 has indeed a different structure

src/main/kotlin/org/elm/ide/toolwindow/ElmCompilerPanel.kt
```diff
@@ -0,0 +1,298 @@
...
+
+    private fun updateSelectionAndCreateOccurenceInfo(): OccurenceNavigator.OccurenceInfo {
+        // update selection
+        messageUI.text = compilerMessages[indexCompilerMessages].messageWithRegion.message
+        errorTableUI.setRowSelectionInterval(indexCompilerMessages, indexCompilerMessages)
+
+        // create occurence info
+        val (virtualFile, document, start) = startFromErrorMessage(VfsUtil.findFile(elmProjects[0].manifestPath, true))
```
klazuka: There is no guarantee that `elmProjects` is non-empty. So this may crash (`elmProjects[0]`).

src/main/kotlin/org/elm/workspace/compiler/ElmBuildAction.kt
```diff
@@ -0,0 +1,95 @@
...
+        if (elmCLI == null) {
+            project.showBalloon("Could not find elm", NotificationType.ERROR, fixAction)
+            return
+        }
+
+        val elmProject = project.elmWorkspace.allProjects.first()
+
+        // find "main" function
+        val clientLocation = LookupClientLocation(project, elmProject)
+        val elements = ElmLookup.findByName<ElmNamedElement>("main", clientLocation)
```
klazuka: I made an attempt at simplifying this whole section of the code. Try this:

```kotlin
// find "main" function
val mainFuncDecl = ElmLookup
        .findByName<ElmFunctionDeclarationLeft>("main", LookupClientLocation(project, elmProject))
        .find {
            val ty = (it.findInference()?.ty as? TyUnion) ?: return@find false
            elmMainTypes.contains(ty.module to ty.name)
        }

if (mainFuncDecl == null) {
    showDialog(project)
    return
}
```

src/main/kotlin/org/elm/workspace/compiler/ElmBuildAction.kt
```diff
@@ -0,0 +1,95 @@
...
+        val mainElements = elements
+                .filter { element -> element.findInference() != null && element.findInference()?.ty is TyUnion }
+                .filter { element -> elmMainTypes.contains(Pair((element.findInference()?.ty as TyUnion).module, (element.findInference()?.ty as TyUnion).name)) }
+
+        if (mainElements.isNotEmpty()) {
+            val path = mainElements[0].containingFile.virtualFile.path
+            try {
```
klazuka: The `try/catch` should just cover the call to `elmCLI.make()` as that is the only part that can throw an `ExecutionException`. Also, Kotlin's try/catch can be used as an expression (e.g. on the right-hand side of the assignment to the local `json` variable), which may be nicer here.

src/main/kotlin/org/elm/workspace/compiler/ElmBuildAction.kt
```diff
@@ -0,0 +1,95 @@
...
+                return
+            }
+        } else {
+            showDialog(project)
+        }
+    }
+
+    private fun showDialog(project: Project) {
+        val statusBar = WindowManager.getInstance().getStatusBar(project)
+        JBPopupFactory.getInstance().createHtmlTextBalloonBuilder("No Type Signature found. Please specify one, so that Elm Build works.", MessageType.ERROR, null).setFadeoutTime(5000).createBalloon().show(RelativePoint.getNorthEastOf(statusBar.component), Balloon.Position.atRight)
```
klazuka: Change to: `Cannot find your Elm app's main entry point. Please make sure that it has a type annotation.`

src/main/kotlin/org/elm/ide/toolwindow/ElmCompilerPanel.kt
clojj: yep... that was a short minded fix

build.gradle
clojj: fort that, yes

clojj:
@klazuka 
please take a look, I think I've gone through all your valuable suggestions.
I am not 100% sure about the way the project's manifest-path is now found via ElmWorkspaceService, but I hope this more like is what you suggested.

klazuka:
Thanks. I will look later today.

src/main/kotlin/org/elm/workspace/compiler/ElmBuildAction.kt
```diff
@@ -0,0 +1,120 @@
...
+                    val ty = (it.findInference()?.ty as? TyUnion) ?: return@find false
+                    elmMainTypes.contains(ty.module to ty.name)
+                }
+
+        if (mainFuncDecl == null) {
+            showDialog(project, "Cannot find your Elm app's main entry point. Please make sure that it has a type annotation.")
+            return
+        }
+
+        val manifestBaseDir = findElmManifestBaseDir(elmProject)
```
klazuka: I would rewrite lines 67-85 as follows in order to make the control flow more consistent:

```kotlin
val manifestBaseDir = VfsUtil.findFile(elmProject.manifestPath, /*refresh*/ true)
if (manifestBaseDir == null) {
    showDialog(project, "Could not find Elm-Project base directory")
    return
}

val json = try {
    elmCLI.make(project, elmProject, mainFuncDecl.containingFile.virtualFile.path).stderr
} catch (e: ExecutionException) {
    project.showBalloon("Invalid path for 'elm' executable", NotificationType.ERROR, fixAction)
    return
}

val messages = if (json.isEmpty()) emptyList() else {
    elmJsonReport.elmToCompilerMessages(json).sortedWith(
            compareBy(
                    { it.name },
                    { it.messageWithRegion.region.start.line },
                    { it.messageWithRegion.region.start.column }
            ))
}
project.messageBus.syncPublisher(ERRORS_TOPIC).update(manifestBaseDir, messages)

// show toolwindow
ToolWindowManager.getInstance(project).getToolWindow("Elm Compiler").show(null)
```

Basically, the idea is that we should follow a consistent pattern of checking for data and returning early if it's not found. It all funnels down to the bottom where the actual work is performed. This makes the code easier to read and easier to understand what it's ultimately trying to do (despite all the error handling).

I also inlined the `findElmManifestBaseDir` function since it wasn't doing much.

src/main/kotlin/org/elm/ide/toolwindow/ElmCompilerPanel.kt
```diff
@@ -0,0 +1,292 @@
...
+import javax.swing.ScrollPaneConstants.HORIZONTAL_SCROLLBAR_NEVER
+import javax.swing.ScrollPaneConstants.VERTICAL_SCROLLBAR_AS_NEEDED
+import javax.swing.border.EmptyBorder
+import javax.swing.table.DefaultTableCellRenderer
+import javax.swing.table.DefaultTableModel
+
+
+class ElmCompilerPanel(private val project: Project, private val contentManager: ContentManager) : SimpleToolWindowPanel(true, false), Disposable, OccurenceNavigator {
+
+    var elmBaseDir: VirtualFile? = null
```
klazuka: yeah, this is much better.

I ran this in the debugger and it's actually referring to the project's `elm.json` file, so the name is misleading. I would rename it to `elmManifestFile` (or change it so that the caller provides the `projectDirPath` rather than the `manifestPath` (the former refers to the _directory_ that contains the `elm.json` file, whereas the latter refers to the `elm.json` file itself)).

clojj:
@klazuka 
comments fixed and 0.18 dropped (that'd be a little bit of wasted energy, agreed)
thx for all your valuable remarks, btw

I also can see, what unit tests would make sense for the new Compiler Toolwindow...

clojj:
Unfortunately, there's sill a problem with main-detection
grafik

This was a newly opened project, initially missing main-signature. Adding the Type signature `Html.Html` did not help

clojj:
I willl check this later...

clojj:
@klazuka 
I just remembered that it is currently not possible to match `TyUnknown` ...which is the inferred type for `Html.Html`
I added this to the message

klazuka:
Ok. I think the problem is that the name resolution code is not resolving the `VirtualDom` module when coming from the `Html` package. I don't have time to fix this right now, so let's just ship it the way you have it with the limitation that it just won't work for toy apps whose main is `Html`.

klazuka:
I think this is ready to go. Let's merge it in. We will still have at least another week before I do another release, so that gives us time to polish things up as needed.

klazuka:
🎉

PR #345, opened 2019-03-08
Assorted improvements to type inference

ajalt:
1. A fix when inferring aliases like `type alias A a = B a` when `B` is an extension record alias
2. Parenthesize parameterized type aliases that are arguments to rendered union types 
3. A new algorithm for type replacement: rather than performing repeated, shallow replacement until a fixed point is reached, we instead perform replacement recursively on the values of the replacement table, allowing all replacements to be completed in a single pass of the ty.
4. Infer unannotated parameters as unconstrained vars (`a`, `b`, etc.) rather than `TyUnknown`. We still don't constrain them, but they are propagated through the body. (`foo a = [a ++ ""]` is inferred as `a -> List String` rather than the correct `String -> List String`, but that's still better than the previous `unknown -> unknown`)

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
+++ b/src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
@@ -909,7 +909,11 @@ private class InferenceScope(
             }
             else -> error("impossible")
         }
-        if (result && ty2.baseTy is TyVar) {
+
+        // We need to unify the base var with the record being assigned for the case where we have
+        // an alias to an extension record constructor. We only do this if the records are
+        // different, since that would incorrectly create a recursive type.
+        if (result && ty2.baseTy is TyVar && ty1.fields != ty2.fields) {
             trackReplacement(ty1, ty2.baseTy, replacements)
         }
         return result
--- a/src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
+++ b/src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
@@ -267,7 +267,7 @@ private class InferenceScope(
                     val lAssignable = requireAssignable(l.start, l.ty, func.parameters[0], l.end, replacements)
                     val rAssignable = requireAssignable(r.start, r.ty, func.parameters[1], r.end, replacements)
                     val ty = when {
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/elm/lang/core/types/TypeRendering.kt
```diff
@@ -26,7 +26,9 @@ fun TyUnion.renderedText(linkify: Boolean, withModule: Boolean): String {
     val type = when {
         parameters.isEmpty() -> name
         else -> parameters.joinToString(" ", prefix = "$name ") {
-            if (it is TyFunction || it is TyUnion && it.parameters.isNotEmpty()) {
+            if (it is TyFunction
+                    || it is TyUnion && it.parameters.isNotEmpty()
+                    || it.alias?.parameters?.isNotEmpty() == true) {
```
klazuka: minor thing: remove the `== true` part

src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
```diff
@@ -624,47 +629,49 @@ private class InferenceScope(
...
-    private fun bindParameters(valueDeclaration: ElmValueDeclaration): Pair<Ty?, Int> {
+    private fun bindParameters(valueDeclaration: ElmValueDeclaration): ParameterBindingResult {
         return when {
             valueDeclaration.functionDeclarationLeft != null -> {
                 bindFunctionDeclarationParameters(valueDeclaration, valueDeclaration.functionDeclarationLeft!!)
             }
             valueDeclaration.pattern != null -> {
                 bindPatternDeclarationParameters(valueDeclaration, valueDeclaration.pattern!!)
-                null to 0
+                ParameterBindingResult.Other(0)
```
klazuka: I like this change a lot (changing from using a pair to using a sealed class). Makes the code easier to read. Plus the improved functionality is nice 😄.

src/main/kotlin/org/elm/lang/core/types/TypeRendering.kt
ajalt: `it.alias?.parameters?.isNotEmpty()` has type `Boolean?`, so the `== true` is required.

PR #351, opened 2019-03-11
Support main functions returning Html.Html in build action

ajalt:
Even when the body of an alias is unresolved, as is the case with `Html.Html`, we still set the alias info on the `TyUnknown`, so we can use that to check the type.

### DIFF
--- a/src/main/kotlin/org/elm/workspace/compiler/ElmBuildAction.kt
+++ b/src/main/kotlin/org/elm/workspace/compiler/ElmBuildAction.kt
@@ -21,6 +21,7 @@ import org.elm.lang.core.lookup.ClientLocation
 import org.elm.lang.core.lookup.ElmLookup
 import org.elm.lang.core.psi.elements.ElmFunctionDeclarationLeft
 import org.elm.lang.core.types.TyUnion
+import org.elm.lang.core.types.TyUnknown
 import org.elm.lang.core.types.findInference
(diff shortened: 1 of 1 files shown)


ajalt:
Note that I'm not able to test this, since the action fails on my Windows machine with `Could not Elm project`  before getting to the type check, but I'm pretty sure this will work.

klazuka:
I had thought about special-casing this, but it seemed brittle just having this hanging out here.

Maybe we could handle this the same way we might handle the implicit, built-in types in 0.18: consolidate the logic for determining if it's a special type in the type inference code. That way if we ever fix it, it will be less likely to be overlooked.

klazuka:
Also, line 62 in `ElmBuildAction` has an error message that says that `Html` type is not supported. So if we remove this limitation, then the error message should also be adjusted.

ajalt:
1. It's up to you; feel free to close this PR if you don't like the approach. I agree that it's not ideal, but this is the first place I've seen that's looking specifically for one of these types. I think there are a lot of aliases to unresolved types, and I didn't want to try to come up with a comprehensive list. The alias info on `TyUnknown` works consistently, and I use it in the documentation provider.

2. I believe I already updated the error message. Did I miss one?

klazuka:
I was wrong about the error message: you already fixed it. You are too good at this stuff! 😄

PR #353, opened 2019-03-12
Build action should work in all cases..

clojj:
Fixed the wording of the error message 
Just use first project from ElmWorkspaceService:
That mechanism "e.getData(...VIRTUAL_FILE)" did not work, if the Build-Button in the compiler-toolwindow is clicked by mouse, even if there is an Elm-file open in the editor.

Right now, there is no "multi-workspace" compiler-mode anyway, so I figure we should just use the first project from ElmWorkspaceService.
(that is,  no multi 'Main.elm'... using 2 or more parameters for 'elm make ...')

### DIFF
--- a/src/main/kotlin/org/elm/workspace/compiler/ElmBuildAction.kt
+++ b/src/main/kotlin/org/elm/workspace/compiler/ElmBuildAction.kt
@@ -4,7 +4,6 @@ import com.intellij.execution.ExecutionException
 import com.intellij.notification.NotificationType
 import com.intellij.openapi.actionSystem.AnAction
 import com.intellij.openapi.actionSystem.AnActionEvent
-import com.intellij.openapi.actionSystem.CommonDataKeys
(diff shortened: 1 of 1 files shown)


clojj:
@klazuka 
Sorry, that's a shortminded fix... I will later do this:
- first check as usual by `getData(...VIRTUAL_FILE)`, if there is any specific file-context, and find project by `ElmWorkspaceService` as before
- if that first method did not find a project, _then_ just use the first project of `allProjects`

Maybe, ideally, the build-action (the button) would be available from the Workspace-Toolwindow, where it can get the 'index' of the currently selected workspace...

What do you think ?

klazuka:
Yep, that’s a good idea!

Last night I started making some changes to your code to make the Kotlin more idiomatic. Please see the ‘kl/cleanup’ branch. One of the changes I made is exactly as you described here in this PR. 

Once the cleanup branch is merged, feel free to open additional PRs.

PR #355, opened 2019-03-12
Clojj/fix color null

clojj:
some markup-chunk have 'null' color (which is a strange enough JSON output from Elm)


clojj:
how you can produce the message, containing a color=null chunk, this is after my fix

color-null-fixed

klazuka:
That is weird. I couldn’t believe it when I saw it in one of your test fixtures, and I was unable to reproduce it. In any case, thanks!

clojj:
@klazuka 
so did you fix it already ? I can't find a commit for this...

klazuka:
Yes. I forgot to push last night before bed.

PR #368, opened 2019-03-20
Update usages of deprecated APIs.

ajalt:
There were several deprecated IntelliJ functions in use that would cause the compiler to emit warnings. All of the APIs have equivalent alternatives.


klazuka:
Thanks. I was planning on doing this once we set 2018.2 as the minimum version, but you beat me to it!

FYI: I’m working on fixing the cache stability warnings in 2019. It’s not super high priority because those exceptions are only enabled in EAP and Internal builds.

PR #374, opened 2019-03-22
add scrollbars to message pane + show compiler target

clojj:


klazuka:
Thank you! 

One thing I noticed while running it is that there's a strange inset on the left-side panel. I'm no expert on Swing, so I have no idea what's going on. IntelliJ does have a [UI inspector](https://www.jetbrains.org/intellij/sdk/docs/reference_guide/internal_actions/internal_uii.html) for debugging these sorts of things.

foo

I'm going to be on vacation until March 31st, so I probably won't be on Github very much.

clojj:
@klazuka 
That must be a default inset, I will try to remove it...

PR #380, opened 2019-03-28
Add reference support for type variables in annotations

ajalt:
This allows you to rename type variables. 

It's also necessary for tracking rigid type variables that reference parent annotations during inference. The fact that type annotations on nested functions can refer to variable in their parent functions was surprising to me. Here's an example of that: https://ellie-app.com/56HqfsFx2Z5a1

Note that the reference semantics for type variables aren't clear with respect to record base identifiers. I ended up treating them like any other type variable. The Elm compiler often crashes when you use them in type annotations.

I also renamed `ElmTypeVariableRef` to `ElmTypeVariable` and merged its anonymous reference implementation into `TypeVariableReference`, since they were almost identical.

Capture


src/main/grammars/ElmParser.bnf
```diff
@@ -224,15 +224,15 @@ TypeRef ::= UpperCaseQID (SingleTypeExpression)+
 
 private SingleTypeExpression ::=
     TypeRefWithoutArgs
-    | TypeVariableRef
+    | TypeVariable
```
klazuka: Nice. The `Ref` suffix is leftover from when I was trying to make a distinction between where types are used vs declared. It was a bad idea since it suggests a relationship to `PsiReference`, so thank you for fixing it here.

PR #384, opened 2019-04-02
Stop treating 'alias' as a restricted keyword

### LINKED ISSUE #378: Incorrect error for records with "alias" as field
I have a record with "alias" as a field. The elm compiler allows it, but the plugin type checking gives the error `LOWER_CASE_IDENTIFIER expected, got 'alias'`.

Thanks for all of the hard work on this project! It makes the development experience so much better.

klazuka:
Fixes #378 

Elm allows code such as:

```elm
foo : { alias : String }
foo = { alias = "" }
```

Previously our lexer was treating these occurrences of `alias` as keywords and so
the expressions failed to parse. But now the lexer will only emit the `alias`
keyword token when it follows a `type` token (and intervening whitespace).


src/main/grammars/ElmLexer.flex
```diff
@@ -121,6 +121,18 @@ ThreeQuotes = \"\"\"
     }
 }
 
+<TYPE_PENDING> {
```
klazuka: Maybe there's a better way to do this in JFlex by looking at token history (or tracking it on our own)?

PR #416, opened 2019-06-04
Fix quick docs for nested functions

ajalt:
Since nested function inference is done in the scope of its top-level
ancestor, we were showing function types for the top level function in
the docs for nested functions.

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
+++ b/src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
@@ -21,6 +21,16 @@ fun PsiElement.findInference(): InferenceResult? {
     return outermostDeclaration(strict = false)?.inference(emptySet())
 }
 
+/**
+ * Find the type of a function declaration.
+ *
+ * This is like [findInference].ty, but handles nested declarations.
+ */
+fun ElmFunctionDeclarationLeft.functionTy(): Ty? {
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
```diff
@@ -21,6 +21,16 @@ fun PsiElement.findInference(): InferenceResult? {
     return outermostDeclaration(strict = false)?.inference(emptySet())
 }
 
+/**
+ * Find the type of a function declaration.
+ *
+ * This is like [findInference].ty, but handles nested declarations.
+ */
+fun ElmFunctionDeclarationLeft.functionTy(): Ty? {
```
klazuka: @ajalt It would read nicer from the caller's perspective if we just used `findTy()` instead of introducing a new method for this case.

The downside is that `findTy()` would have to have a fallback to handle this case. But it seems more preferrable than putting the burden on the caller to understand how the type inference data structures work.

klazuka: Hmm... I guess the strange thing here is that this is on `ElmFunctionDeclarationLeft`, which is really a fragment. It doesn't truly have a type.

This wart would go away when I finally bite the bullet and flatten out `InternalValueDeclarationLeft` so that each rule becomes its own top-level declaration element. I think that would be an improvement, and would make this less gross.

PR #417, opened 2019-06-05
Add MakeAnnotationIntention

ajalt:
This inspection triggers on function declarations (left of the =) that don't have an annotation. Now that we properly constrain type variables with the completion of the unification work, we can generate a type annotation without breaking existing code.

Additionally:

- Fix MakeDeclarationIntention when used on nested declarations
- Remove `PsiElement.guessIndent` in favor of `DocumentUtil.getIndent` (it's better to get the actual indentation rather than guess)
- A few assorted typo and lint fixes


src/main/kotlin/org/elm/ide/intentions/MakeAnnotationIntention.kt
```diff
@@ -0,0 +1,49 @@
...
+
+        val ty = declaration.findTy() ?: return null
+
+        return Context(fdl, declaration, ty)
+    }
+
+    override fun invoke(project: Project, editor: Editor, context: Context) {
+        val (fdl, valueDeclaration, ty) = context
+        val indent = editor.getIndent(valueDeclaration.startOffset)
+        val code = "${fdl.name} : ${ty.renderedText(false, false).replace("→", "->")}\n$indent${valueDeclaration.text}"
```
klazuka: I think this comes out a little bit nicer if you do an insert instead of replace:

```kotlin
val code = "${fdl.name} : ${ty.renderedText(false, false).replace("→", "->")}\n$indent"
project.runWriteCommandAction {
        editor.document.insertString(valueDeclaration.startOffset, code)
}
```

src/main/kotlin/org/elm/ide/intentions/MakeDeclarationIntention.kt
```diff
@@ -49,7 +52,8 @@ class MakeDeclarationIntention : ElmAtCaretIntentionActionBase<MakeDeclarationIn
         val anchor = typeAnnotation.nextLeaves
                 .takeWhile { !it.text.contains('\n') }
                 .lastOrNull() ?: typeAnnotation
-        typeAnnotation.parent.addAfter(factory.createFreshLine(), anchor)
+        val indent = editor.getIndent(typeAnnotation.startOffset)
+        typeAnnotation.parent.addAfter(factory.createWhitespace("\n$indent"), anchor)
```
klazuka: Thank you for fixing my bugs 🙏

src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
```diff
@@ -21,18 +21,21 @@ fun PsiElement.findInference(): InferenceResult? {
     return outermostDeclaration(strict = false)?.inference(emptySet())
 }
 
-/**
- * Find the type of a function declaration.
- *
- * This is like [findInference].ty, but handles nested declarations.
- */
-fun ElmFunctionDeclarationLeft.functionTy(): Ty? {
```
klazuka: Nice! It's a bit of a lie, but it seems worth it.

klazuka:
When generating a type annotation, there's a bug whenever a type in the annotation needs to be qualified by the imported module. The generated annotation is missing the qualifier, and so the reference is unresolved.

See https://github.com/rtfeldman/elm-spa-example/blob/17a3398623f9e538f14f5b0d039fd62b3beae934/src/Page/Login.elm#L293

The intention generates `login : TrimmedForm -> Request Viewer`, but `Request` needs to be qualified by `Http` due to how it has been imported.

klazuka:
When generating a type annotation involving a record, it is a bit too general, favoring a record type literal over an alias.

For instance, try generating a type annotation for `viewForm` in elm-spa-example:
https://github.com/rtfeldman/elm-spa-example/blob/17a3398623f9e538f14f5b0d039fd62b3beae934/src/Page/Login.elm#L120

It generates:
```
viewForm : { a | email : String, password : String } -> Html Msg
```

But I would expect it to generate `viewForm : Form -> Html Msg`, especially since all the record fields are used.

This is also a problem in functions like `update` where it generates a huge, gross type annotation filled with record type literals instead of the expected type alias. I realize that this case may be difficult to handle, especially if the function uses a subset of the record fields defined in the type alias. But maybe it would be worth putting in a hack?

klazuka:
> When generating a type annotation, there's a bug whenever a type in the annotation needs to be qualified by the imported module. The generated annotation is missing the qualifier, and so the reference is unresolved.

As discussed at lunch, let's ignore this and instead deal with it by implementing #419

PR #420, opened 2019-06-06
Add 'introduce variable' refactoring

klazuka:
Suppose you have code like:
```elm
f x =
    if x > 0 && x < 100 then Just x else Nothing
```
... and you want to factor-out the `if`'s boolean predicate. You can now put the cursor anywhere in the boolean expression, press Cmd-Option-V (Ctrl-Alt-V on Windows/Linux), and the plugin will move the expression into a `let` like so:
```elm
f x =
    let
         isValid = x > 0 && x < 100
    in 
    if isValid then Just x else Nothing
```

This works for all expressions. And if you already have a `let`, it will just add another declaration to that `let`.

The plugin will automatically suggest names for the introduced variable based on the expression, and it will automatically invoke in-place rename so that you can provide a better name if desired.

## Implementation Notes

I followed intellij-rust very closely, but the code generation was trickier (partly due to Elm's whitespace-sensitive parsing and partly due to how intrusive adding a `let` is). It could benefit from some cleanup, but it will have to suffice for now.

I punted on the "replace all occurrences" feature as the code is already complicated enough as-is. I will implement it later.

### DIFF
--- a/src/main/kotlin/org/elm/ide/refactoring/ElmIntroduceVariableHandler.kt
+++ b/src/main/kotlin/org/elm/ide/refactoring/ElmIntroduceVariableHandler.kt
@@ -75,10 +75,12 @@ class ElmIntroduceVariableHandler : RefactoringActionHandler {
         // find the nearest location where a let/in would be
         var current: PsiElement? = chosenExpr
         while (current != null) {
-            when (current) {
-                is ElmValueDeclaration -> return current.expression
-                is ElmCaseOfBranch -> return current.expression
-                is ElmAnonymousFunctionExpr -> return current.expression
+            when {
+                current is ElmValueDeclaration -> return current.expression
+                current is ElmCaseOfBranch -> return current.expression
+                current is ElmAnonymousFunctionExpr -> {
+                    if (current.expression in chosenExpr.ancestors) return current.expression
+                }
             }
             current = current.parent
         }
--- a/src/main/kotlin/org/elm/ide/refactoring/ElmIntroduceVariableHandler.kt
+++ b/src/main/kotlin/org/elm/ide/refactoring/ElmIntroduceVariableHandler.kt
@@ -146,7 +146,8 @@ private class ExpressionReplacer(
         val file = letExpr.elmFile
         val anchor = letExpr.valueDeclarationList.last()
         val indent = DocumentUtil.getIndent(editor.document, anchor.startOffset)
-        val indentedDeclExpr = chosenExpr.text.lines().joinToString("\n") { "${indent}    ${it.trimStart()}" }
+        val indentedDeclExpr = chosenExpr.textWithNormalizedIndents.lines()
+                .joinToString("\n") { "$indent    $it" }
         val textToInsert = "\n\n$indent${identifier.text} =\n$indentedDeclExpr"
         val offsetOfNewDecl = anchor.endOffset + textToInsert.indexOf(identifier.text)
 
--- a/src/main/kotlin/org/elm/ide/refactoring/ElmIntroduceVariableHandler.kt
+++ b/src/main/kotlin/org/elm/ide/refactoring/ElmIntroduceVariableHandler.kt
@@ -19,6 +19,7 @@ import org.elm.lang.core.psi.elements.ElmAnonymousFunctionExpr
 import org.elm.lang.core.psi.elements.ElmCaseOfBranch
 import org.elm.lang.core.psi.elements.ElmLetInExpr
 import org.elm.lang.core.psi.elements.ElmValueDeclaration
+import org.elm.lang.core.textWithNormalizedIndents
 import org.elm.openapiext.runWriteCommandAction
 
 class ElmIntroduceVariableHandler : RefactoringActionHandler {
@@ -96,7 +97,6 @@ class ElmIntroduceVariableHandler : RefactoringActionHandler {
  * Manages the replacement of an expression with the appropriate let/in expr
  *
  * This class exists solely to make the calling code nicer by consolidating shared parameters.
- * (poor-man's partially applied functions)
  */
 private class ExpressionReplacer(
         private val project: Project,
@@ -107,20 +107,27 @@ private class ExpressionReplacer(
     private val suggestedNames = chosenExpr.suggestedNames()
     private val identifier = psiFactory.createLowerCaseIdentifier(suggestedNames.default)
 
+    /*
+    NOTE If you ever have to make changes to how the Elm code is generated, especially
+         with respect to indentation, you will almost certainly want to rewrite this.
+         I have decided to go with dirty, ad-hoc string manipulation in the interest
+         of shipping this feature quickly. The right thing to do is probably to
+         integrate with IntelliJ's whitespace formatting system.
+    */
+
+
     /**
      * Wrap the existing function body in a `let` expression
      */
     fun introduceLet(elementToReplace: PsiElement) {
         val indent = DocumentUtil.getIndent(editor.document, elementToReplace.startOffset).toString()
-        val exprText = chosenExpr.text
-        val declText = "${identifier.text} =\n    $exprText"
-
+        val newDeclBodyText = chosenExpr.textWithNormalizedIndents
         val newIdentifierElement = project.runWriteCommandAction {
             val newLetExpr = if (elementToReplace !== chosenExpr) {
                 chosenExpr.replace(identifier)
-                psiFactory.createLetInWrapper(indent, declText, elementToReplace.text)
+                psiFactory.createLetInWrapper(indent, identifier.text, newDeclBodyText, elementToReplace.text)
             } else {
-                psiFactory.createLetInWrapper(indent, declText, identifier.text)
+                psiFactory.createLetInWrapper(indent, identifier.text, newDeclBodyText, identifier.text)
             }
             val newLetElement = elementToReplace.replace(newLetExpr) as ElmLetInExpr
             moveEditorToNameElement(editor, newLetElement.valueDeclarationList.first())
@@ -139,7 +146,8 @@ private class ExpressionReplacer(
         val file = letExpr.elmFile
         val anchor = letExpr.valueDeclarationList.last()
         val indent = DocumentUtil.getIndent(editor.document, anchor.startOffset)
-        val textToInsert = "\n\n$indent${identifier.text} =\n$indent    ${chosenExpr.text}"
+        val indentedDeclExpr = chosenExpr.text.lines().joinToString("\n") { "${indent}    ${it.trimStart()}" }
+        val textToInsert = "\n\n$indent${identifier.text} =\n$indentedDeclExpr"
         val offsetOfNewDecl = anchor.endOffset + textToInsert.indexOf(identifier.text)
 
         /*
--- a/src/main/kotlin/org/elm/ide/refactoring/ElmIntroduceVariableHandler.kt
+++ b/src/main/kotlin/org/elm/ide/refactoring/ElmIntroduceVariableHandler.kt
@@ -66,7 +66,7 @@ class ElmIntroduceVariableHandler : RefactoringActionHandler {
         val replacer = ExpressionReplacer(project, editor, chosenExpr)
         val anchor = findAnchor(chosenExpr) ?: error("could not find a place to introduce variable")
         when {
-            anchor is ElmLetInExpr -> replacer.extendExistingLet(anchor)
+            anchor is ElmLetInExpr && anchor !== chosenExpr -> replacer.extendExistingLet(anchor)
             else -> replacer.introduceLet(anchor)
         }
     }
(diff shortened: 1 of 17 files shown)


src/main/kotlin/org/elm/ide/refactoring/ElmIntroduceVariableHandler.kt
```diff
@@ -0,0 +1,188 @@
...
+
+class ElmIntroduceVariableHandler : RefactoringActionHandler {
+
+    override fun invoke(project: Project, editor: Editor, file: PsiFile, dataContext: DataContext?) {
+        if (file !is ElmFile) return
+        val exprs = findCandidateExpressions(editor, file)
+        when (exprs.size) {
+            0 -> {
+                val message = RefactoringBundle.message(if (editor.selectionModel.hasSelection())
+                    "selected.block.should.represent.an.expression"
```
klazuka: localized string keys provided by IntelliJ

src/main/kotlin/org/elm/ide/refactoring/ElmIntroduceVariableHandler.kt
```diff
@@ -0,0 +1,188 @@
...
+        val replacer = ExpressionReplacer(project, editor, chosenExpr)
+        val anchor = findAnchor(chosenExpr) ?: error("could not find a place to introduce variable")
+        when {
+            anchor is ElmLetInExpr -> replacer.extendExistingLet(anchor)
+            else -> replacer.introduceLet(anchor)
+        }
+    }
+
+    private fun findAnchor(chosenExpr: ElmExpressionTag): ElmPsiElement? {
+        // find the nearest location where a let/in would be
```
klazuka: In the future we may want to allow the user to choose which scope they would like to add it to instead of assuming that it's always the nearest scope.

src/main/kotlin/org/elm/ide/utils/SearchByOffset.kt
```diff
@@ -0,0 +1,107 @@
...
+import com.intellij.psi.PsiFile
+import com.intellij.psi.PsiWhiteSpace
+import com.intellij.psi.util.PsiTreeUtil
+import org.elm.lang.core.psi.*
+
+/**
+ * Find an [ElmExpressionTag] element at [offset], being smart about whether
+ * we look to the left or the right of the caret offset.
+ */
+fun findExpressionAtCaret(file: ElmFile, offset: Int): ElmExpressionTag? {
```
klazuka: All of the code in this file came from intellij-rust with minor adjustments for Elm Psi.

src/main/kotlin/org/elm/lang/core/ElmTextUtils.kt
```diff
@@ -0,0 +1,15 @@
...
+ * Convert a string so that its first character is guaranteed to be lowercase.
+ * This is necessary in some parts of Elm's syntax (e.g. a function parameter).
+ *
+ * If the receiver consists of all uppercase letters, the entire thing will be made
+ * lowercase (because "uuid" is a far more sensible transformation of "UUID" than "uUID").
+ */
+fun String.toElmLowerId(): String =
+        when {
+            isEmpty() -> ""
+            all { it.isUpperCase() } -> toLowerCase()
```
klazuka: This fixes a long-standing bug when generating a func decl from a type annotation. If you had `f : UUID -> String` and you invoked the 'make declaration' intention, it would generate `f uUID = ...`, which is lame. Now it will generate `f uuid = ...`

src/main/kotlin/org/elm/lang/core/psi/ElmPsiFactory.kt
```diff
@@ -140,6 +140,18 @@ class ElmPsiFactory(private val project: Project) {
                     ?.childOfType<ElmCaseOfExpr>()?.branches
                     ?: error("Failed to create case of branches from $patterns")
 
+    fun createLetInWrapper(indent: String, newDeclText: String, bodyText: String): ElmLetInExpr =
+            """
+foo =
```
klazuka: super gross. And now that I'm looking at this, I'm going to try simplifying it.

src/main/kotlin/org/elm/lang/core/psi/elements/ElmFunctionDeclarationLeft.kt
```diff
@@ -69,4 +64,24 @@ class ElmFunctionDeclarationLeft : ElmStubbedNamedElementImpl<ElmFunctionDeclara
             val p = parent
             return p is ElmValueDeclaration && p.isTopLevel
         }
+
+    override fun getUseScope(): SearchScope {
```
klazuka: This was necessary so that we can trigger the inplace rename refactoring automatically, but it also provides a nice speed bonus for anyone who wants to rename an inner decl in a `let`.

src/main/resources/META-INF/plugin.xml
```diff
@@ -156,7 +156,7 @@
     </change-notes>
 
     <!-- please see http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/build_number_ranges.html for description -->
-    <idea-version since-build="182.3684.2"/>
+    <idea-version since-build="183.3975.18"/>
```
klazuka: `DocumentUtil.getIndent` was introduced in 2018.3

ajalt:
🔥 This is a super helpful feature. I know it must have been really tough to get working with the way Elm handles whitespace. Here's a few bugs I found while playing around with it:

### Case branch patterns

```elm
example x =
    case x of 
        [extractThis, b] -> ()
```

Extracting `extractThis` produces:

```elm
example x =
    let
        x1 = case x of
        [extractThis, b] -> ()
    in
    x1
```

### Lambdas inside parentheses

```elm
example2 =
     (\{ extractTheLambdaButNotTheParentheses } -> ())
```

Extracting the lambda expression but not the enclosing parentheses:

```elm
example2 =
     (f)
```

### Entire case expressions

This one just produces incorect indentation:

```elm
example3 =
    case () of
        () -> ()
```

Extracting the case expression:

```elm
example3 =
    let
        x = case () of
        () -> ()
    in
    x
```

### Records

```elm
example4 =
    { field = 1
    , field2 = 2
    }
```

extracting the record produces:

```elm
example4 =
    let
        x = { field = 1
```

### Multi-line expressions

```elm
example5 =
    1
    +
    2
```

Extracting the `+`   expression:

```elm
example5 =
    let
        x = 1
```

Also:

```elm
example6 =
    1
        + 2
```

Extracting the `+` expression:

```elm
example6 =
    let
        x = 1
        + 2
    in
    x
```

The `+ 2` line needs to be indented more than the `x = 1` line to produce valid code.

### Nested let expressions:

```elm
example7 =
    let
        x = 1
    in
    let
        y = 2
    in
    y
```

Extracting the usage of `y`:

```elm
example7 =
    let
        x = 1

        x1 = y
    in
    let
        y = 2
    in
    x1
```

### Extracting entire let-in

```elm
module Empty exposing (example8)

example8 =
    let
        x = 1
    in
    2
```

Manually select from the start of `let` to the end of the `2`:

```elm
module Empty exposing (exampl

        x1 = let
        x =
            1
    in
    2e8)

example8 =
    x1
```

klazuka:
Drat! I knew about some of these cases where the indent doesn't look right, but I figured people would just run elm-format to fix things up (lame, but shrug).

I didn't know about the other cases where it generates completely malformed code. Thanks for providing examples.

klazuka:
@ajalt I fixed everything except example 7 (it doesn't seem worth the effort, especially when I'm busy moving across the country).

klazuka:
The Windows CI build on Travis busted with some internal Gradle error. Not sure what's going on there. But in any case, I'm somewhat worried that this PR will not work when using Windows line-endings. It appears that there are a few places where we insert just "\n". Maybe IntelliJ automatically handles it for you?

src/main/kotlin/org/elm/ide/refactoring/extractExpressionUI.kt
```diff
@@ -0,0 +1,40 @@
...
+private val <T> ((T) -> Unit).asPass: Pass<T>
+    get() = object : Pass<T>() {
+        override fun pass(t: T) = this@asPass(t)
+    }
+
+interface ExtractExpressionUi {
+    fun chooseTarget(exprs: List<ElmExpressionTag>): ElmExpressionTag
+}
+
+var MOCK: ExtractExpressionUi? = null
```
ajalt: This can be private

src/main/kotlin/org/elm/ide/refactoring/extractExpressionUI.kt
```diff
@@ -0,0 +1,40 @@
...
+@TestOnly
+fun withMockTargetExpressionChooser(mock: ExtractExpressionUi, f: () -> Unit) {
```
ajalt: nbd: You could potentially make the callsite simpler if you get rid of the `ExtractExpressionUi` interface and make this `mock: (List<ElmExpressionTag>) -> ElmExpressionTag`. 

Which would let you do

```kotlin
withMockTargetExpressionChooser(
        mock = { exprs ->
            //...
        }
) {
   // ...
}
```

src/main/kotlin/org/elm/ide/refactoring/extractExpressionUI.kt
```diff
@@ -0,0 +1,40 @@
...
+
+fun showExpressionChooser(
+        editor: Editor,
+        exprs: List<ElmExpressionTag>,
+        callback: (ElmExpressionTag) -> Unit
+) {
+    if (isUnitTestMode) {
+        callback(MOCK!!.chooseTarget(exprs))
+    } else {
+        IntroduceTargetChooser.showChooser(editor, exprs, callback.asPass) { it.text }
```
ajalt: nit: `asPass` is pretty gnarly for something only called once. It might be more readable to create the object manually:

```kotlin
val pass = object : Pass<ElmExpressionTag>() {
    override fun pass(t: ElmExpressionTag) = callback(t)
}
IntroduceTargetChooser.showChooser(editor, exprs, pass) { it.text }
```

src/main/kotlin/org/elm/ide/refactoring/nameSuggestions.kt
```diff
@@ -0,0 +1,70 @@
...
+            // suggest the last field in a record field access chain (e.g. "title" in expr `model.currentPage.title`)
+            names.addName(this.lowerCaseIdentifier?.text)
+        }
+    }
+
+    // if present, suggest type alias (useful for records which often have a descriptive alias name)
+    ty?.alias?.name?.let { names.addName(it) }
+
+    // suggest based on common types
+    when (ty) {
```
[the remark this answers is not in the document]
klazuka: Thanks. I did everything except special-case `TyFunction`. I don't see what you're getting at by checking the return type. Are you assuming that the user is going to apply the args within the function body? My assumption is that they would instead add parameters to the function declaration in the `let` (it would be nice to do that automatically, but not now).

src/main/kotlin/org/elm/ide/refactoring/extractExpressionUI.kt
klazuka: This interface will eventually grow to include additional hooks (choose anchor scope and the option to replace-all duplicate code expressions)

src/main/kotlin/org/elm/ide/refactoring/extractExpressionUI.kt
klazuka: This util came from intellij-rust. Once the interface grows to include more hooks, it will make more sense to have this thing extracted out. So I'm going to leave it.

src/main/kotlin/org/elm/ide/refactoring/ElmIntroduceVariableHandler.kt
```diff
@@ -0,0 +1,199 @@
...
+        when {
+            anchor is ElmLetInExpr && anchor !== chosenExpr -> replacer.extendExistingLet(anchor)
+            else -> replacer.introduceLet(anchor)
+        }
+    }
+
+    private fun findAnchor(chosenExpr: ElmExpressionTag): ElmPsiElement? {
+        // find the nearest location where a let/in would be
+        var current: PsiElement? = chosenExpr
+        while (current != null) {
```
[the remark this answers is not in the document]
klazuka: I fixed the example7 case (thanks for the simple solution), and I totally agree with you that it would be nice to special-case `ElmValueDeclaration` so that we try to re-use an existing `let` as much as possible. The problem is that if the decl is a function with declared parameters, then it introduces new names in the lexical scope, in which case it should not be lifted up to reuse the existing `let`.

We could also handle that case, but I think this would best be handled if/when I add support for the user to choose an anchor scope when multiple scopes are possible.

klazuka:
Thanks for finding all these bugs. I regret putting-off integrating with IntelliJ's whitespace formatting system. I will attempt to fix the single-line expression examples, but "dedent" may have to wait until we have more control over how the code is formatted.

src/main/kotlin/org/elm/ide/refactoring/nameSuggestions.kt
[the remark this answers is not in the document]
ajalt: My thought was that if you extract, say, a partial application of your `Model` constructor, it would make sense to call the function `model`, since that's the value of the extracted function after you add parameters. Regardless, people probably don't use the default names often, so it doesn't matter either way.

klazuka:
It's time to get this thing landed. I believe all of the bugs you found were in code that was formatted strangely. I'm comfortable shipping this as-is with the expectation that people will be using it together with `elm-format`. It's not great, but I think until we have a way to do the formatting automatically (or look into whether the Introduce Variable refactoring could shell-out to external process), this is a reasonable tradeoff.

PR #426, opened 2019-07-08
Upgrade Kotlin and the Gradle-IntelliJ plugin

klazuka:


klazuka:
The new version of the gradle-intellij plugin fixes the broken build on Windows Travis CI.

PR #427, opened 2019-07-08
Upgrade IntelliJ platform to 2019.1.2

klazuka:


klazuka:
@ajalt infinite type recursion checker appears to be broken on IntelliJ 2019. I noticed it when trying to upgrade to 2019.2 EAP, but it also appears to be broken in 2019.1. I haven't traced through the code yet, but I thought if you had a few minutes, you could probably figure it out faster than me. Any ideas?

PR #449, opened 2019-07-25
Check `elm --version` in bg (and cache it)

### LINKED ISSUE #447: Do not synchronously query `elm --version` while ReadAction held
When running an internal build using IDEA 2019.2, the following helpful exception is thrown.
```
java.lang.Throwable: Synchronous execution under ReadAction: /usr/local/bin/elm --version
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:145)
	at com.intellij.execution.process.OSProcessHandler.checkEdtAndReadAction(OSProcessHandler.java:117)
	at com.intellij.execution.process.OSProcessHandler.waitFor(OSProcessHandler.java:62)
	at com.intellij.execution.process.CapturingProcessRunner.runProcess(CapturingProcessRunner.java:53)
	at com.intellij.execution.process.CapturingProcessHandler.runProcess(CapturingProcessHandler.java:69)
	at org.elm.openapiext.SubprocessesKt.execute(Subprocesses.kt:115)
	at org.elm.openapiext.SubprocessesKt.execute$default(Subprocesses.kt:97)
	at org.elm.workspace.commandLineTools.ElmCLI.queryVersion(ElmCLI.kt:42)
	at org.elm.ide.notifications.ElmNeedsConfigNotificationProvider.createNotificationPanel(ElmNeedsConfigNotificationProvider.kt:62)
	at org.elm.ide.notifications.ElmNeedsConfigNotificationProvider.createNotificationPanel(ElmNeedsConfigNotificationProvider.kt:22)
	at com.intellij.ui.EditorNotifications$Provider.createNotificationPanel(EditorNotifications.java:37)
```

Since we are being called with the read action already held, our only choice is to do the work on a background thread, cache it, and then ask IntelliJ to refresh the editor notification panel, this time using the cached version. And figure out when to invalidate the cache.

klazuka:
Fixes #447

### DIFF
--- a/src/main/kotlin/org/elm/ide/notifications/ElmNeedsConfigNotificationProvider.kt
+++ b/src/main/kotlin/org/elm/ide/notifications/ElmNeedsConfigNotificationProvider.kt
@@ -1,11 +1,13 @@
 package org.elm.ide.notifications
 
+import com.intellij.execution.process.ProcessIOExecutorService
 import com.intellij.notification.NotificationType
+import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.diagnostic.logger
 import com.intellij.openapi.fileEditor.FileEditor
 import com.intellij.openapi.fileEditor.OpenFileDescriptor
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.Key
-import com.intellij.openapi.util.SimpleModificationTracker
 import com.intellij.openapi.vfs.LocalFileSystem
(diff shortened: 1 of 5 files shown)


klazuka:
This *was* being called on a background thread. But the problem was that we were holding the read lock. I could assert both, but then we have to make a special exception for the tests which use the main (EDT) thread. I think I'd rather rely on discipline than a false sense of security.

PR #450, opened 2019-07-26
Some syntax highlighting improvements

klazuka:
- fixed a bug on record field types (they were getting the union variant color)
- cleaned up the names of things
- improved the union variant color in the Default/Light scheme
- simplified the user-facing choices

I also laid the groundwork for highlighting module qualifiers in SHA e692b91, but I had a hard time finding a color that worked well in all the contexts. So I punted. The problem is that the module qualifier color needs to work adjacent to 3 different things, each with their own color
- value identifier
- type expression
- union variant 

I'm sure that a good solution can be achieved, but it's just going to take more time. FWIW, here's the color scheme tool I was using in case anyone wants to take a stab at it: https://coolors.co/660e7a-127b7e-004e64-ffffff-000080

### DIFF
--- a/src/main/kotlin/org/elm/ide/highlight/ElmSyntaxHighlightAnnotator.kt
+++ b/src/main/kotlin/org/elm/ide/highlight/ElmSyntaxHighlightAnnotator.kt
@@ -12,34 +12,40 @@ import org.elm.lang.core.psi.elements.*
 class ElmSyntaxHighlightAnnotator : Annotator {
 
     override fun annotate(element: PsiElement, holder: AnnotationHolder) {
+        val h = Highlighter(holder)
         when (element) {
-            is ElmValueDeclaration -> highlightValueDeclaration(holder, element)
-            is ElmTypeAnnotation -> highlightTypeAnnotation(holder, element)
-            is ElmUpperCaseQID -> highlightUpperCaseQID(holder, element)
-            is ElmField -> highlightField(holder, element.lowerCaseIdentifier)
-            is ElmFieldType -> highlightField(holder, element.lowerCaseIdentifier)
-            is ElmFieldAccessExpr -> highlightField(holder, element.lowerCaseIdentifier)
-            is ElmFieldAccessorFunctionExpr -> highlightFieldAccessorFunction(holder, element)
-            is ElmUnionVariant -> highlightTypeConstructor(holder, element.upperCaseIdentifier)
-            is ElmTypeVariable -> highlightTypeExpr(holder, element)
-            is ElmLowerTypeName -> highlightTypeExpr(holder, element)
+            is ElmValueDeclaration -> h.valueDeclaration(element)
+            is ElmTypeAnnotation -> h.typeAnnotation(element)
+            is ElmUpperCaseQID -> h.upperCaseQID(element)
+            is ElmField -> h.field(element.lowerCaseIdentifier)
+            is ElmFieldType -> h.field(element.lowerCaseIdentifier)
+            is ElmFieldAccessExpr -> h.field(element.lowerCaseIdentifier)
+            is ElmFieldAccessorFunctionExpr -> h.fieldAccessorFunction(element)
+            is ElmUnionVariant -> h.unionVariant(element.upperCaseIdentifier)
+            is ElmTypeVariable -> h.typeExpr(element)
+            is ElmLowerTypeName -> h.typeExpr(element)
         }
     }
+}
 
-    private fun highlightTypeExpr(holder: AnnotationHolder, element: PsiElement) {
-        highlightElement(holder, element, ElmColor.TYPE_EXPR)
+
+@Suppress("EXPERIMENTAL_FEATURE_WARNING")
+inline class Highlighter(private val holder: AnnotationHolder) {
+
+    fun typeExpr(element: PsiElement) {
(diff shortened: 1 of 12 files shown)


src/main/kotlin/org/elm/ide/highlight/ElmSyntaxHighlightAnnotator.kt
```diff
@@ -79,6 +79,7 @@ class ElmSyntaxHighlightAnnotator : Annotator {
     }
 
     private fun highlightElement(holder: AnnotationHolder, element: PsiElement, color: ElmColor) {
+        //println("Highlighting %-60s \"%-20s\" with %s".format(element, element.text, color))
```
ajalt: Did you mean to leave this in?

src/main/kotlin/org/elm/ide/highlight/ElmSyntaxHighlightAnnotator.kt
```diff
@@ -12,74 +12,73 @@ import org.elm.lang.core.psi.elements.*
 class ElmSyntaxHighlightAnnotator : Annotator {
 
     override fun annotate(element: PsiElement, holder: AnnotationHolder) {
+        val h = Highlighter(holder)
```
klazuka: @ajalt what do you think about this? I wanted to partially apply the holder so that we don't have to pass it around so much. But this function will be called for every element in the Psi tree, so I assume that I should be careful about unnecessary heap allocations.

I don't know much about reading Kotlin/JVM bytecode, but it looks good:
```
   L1
    LINENUMBER 15 L1
    ALOAD 2
    INVOKESTATIC org/elm/ide/highlight/Highlighter.constructor-impl (Lcom/intellij/lang/annotation/AnnotationHolder;)Lcom/intellij/lang/annotation/AnnotationHolder;
    ASTORE 3
   L2
    LINENUMBER 16 L2
    ALOAD 1
    ASTORE 4
   L3
    LINENUMBER 17 L3
    ALOAD 4
    INSTANCEOF org/elm/lang/core/psi/elements/ElmValueDeclaration
    IFEQ L4
   L5
    ALOAD 3
    ALOAD 1
    CHECKCAST org/elm/lang/core/psi/elements/ElmValueDeclaration
    INVOKESTATIC org/elm/ide/highlight/Highlighter.valueDeclaration-impl (Lcom/intellij/lang/annotation/AnnotationHolder;Lorg/elm/lang/core/psi/elements/ElmValueDeclaration;)V
    GOTO L6
   L4
    LINENUMBER 18 L4
    ALOAD 4
    INSTANCEOF org/elm/lang/core/psi/elements/ElmTypeAnnotation
    IFEQ L7
   L8
    ALOAD 3
    ALOAD 1
    CHECKCAST org/elm/lang/core/psi/elements/ElmTypeAnnotation
    INVOKESTATIC org/elm/ide/highlight/Highlighter.typeAnnotation-impl (Lcom/intellij/lang/annotation/AnnotationHolder;Lorg/elm/lang/core/psi/elements/ElmTypeAnnotation;)V
    GOTO L6
```

Am I interpreting this correctly?

ajalt: This is calling static functions with the annotation holder as the first argument, so no extra heap allocations. You could go one small step further to avoid the `h.`s and move this function into the `Hightlighter`, and call it like `Highlighter(holder).highlight(element)`. Then all the other `Highlighter` methods could be private.

ajalt:
It looks like the entire function declaration left is being highlighted like it the function name:
Capture

klazuka:
> I'm still of the opinion that separate module colors is too noisy. The goal is that the qualifiers have less visual importance than the thing they're qualifying, right? I think giving them a separate color probably has the opposite effect. Maybe we could leave them the same color as the element they're attached to, but make them non-bold or italic or a shade less saturated?

That's what I thought too. I tried italics, and it looks good in small quantities, but if you use qualified refs a lot, it gets too messy. Choosing a less saturated color is basically what @chadtech was doing in his example, but we would need to specify a separate color for each of the 3 cases above (because unlike most things that you want to color, this is tacked on as a prefix to another token, so they have to work well together).

src/main/kotlin/org/elm/ide/highlight/ElmSyntaxHighlightAnnotator.kt
```diff
@@ -12,73 +12,73 @@ import org.elm.lang.core.psi.elements.*
...
+            is ElmTypeVariable -> h.typeExpr(element)
+            is ElmLowerTypeName -> h.typeExpr(element)
         }
     }
+}
+
 
-    private fun highlightTypeExpr(holder: AnnotationHolder, element: PsiElement) {
-        highlightElement(holder, element, ElmColor.TYPE_ANNOTATION_SIGNATURE_TYPES)
+@Suppress("EXPERIMENTAL_FEATURE_WARNING")
```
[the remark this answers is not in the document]
klazuka: at least this time I bothered with suppressing the warning. Those KeyPathRedactor warnings had been going off in PowerLift for a looong time 😃 

But point taken. I will fix it 😃

klazuka:
> It looks like the entire function declaration left is being highlighted like it the function name

I know, I'm fixing it right now

PR #453, opened 2019-07-31
Setup CI with Azure Pipelines

klazuka:


klazuka:
Travis CI's support for Windows was terrible, and it hasn't improved at all since it was released. This appears to be better.

PR #454, opened 2019-08-01
Add intention to generate JSON encoder and decoder from a type annotation

### LINKED ISSUE #424: Generate JSON decoder/encoder
Apparently Elmjutsu does this for Atom. I assume it’s a best-effort thing where the record field names correspond directly to the JSON property names. 

I’m not sure how to surface this in the UI in a discoverable way. Needs investigation.

ajalt:
This PR adds "Generate encoder" and "Generate decoder" intentions that will generate functions for type annotations that return JSON encoders or decoders.

Some notable features of the generated code:

- Code is always syntactically valid, although it may try to import type constructors that aren't exposed in their modules. It will generate `Debug.todo` messages for types it doesn't know how to generate code for.
- It generates separate functions for each type and type alias, rather than inlining them. 
- It will use existing encoders/decoders from any module imported in the current scope, so e.g you could manually write the encoder for an unusual type, and the generated code will call that function to encode record fields of that type.
- It properly qualifies all references, so even if you import a type with an alias or expose it, the generated code will be correct.
- It uses `Json.Decode.Pipeline` for decoding records. That module isn't part of the standard library, but it's much better than the built in `mapN` functions.

---

before

---

after

Fixes #424


src/main/kotlin/org/elm/ide/intentions/AddImportIntention.kt
```diff
@@ -268,3 +206,65 @@ data class Candidate(
                     null
     }
 }
+
+object ImportAdder {
```
ajalt: I moved this code unchanged into these static functions. Let me know if I should move it to a separate file, or structure it some other way.

src/main/kotlin/org/elm/ide/intentions/TyFunctionGenerator.kt
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /** Prefix [name], defined in [module], with the necessary qualifier */
+    protected fun qual(module: String, name: String) = "${qualifierFor(Ref(module, name))}$name"
+
+    /** Code to insert after the type annotation */
+    abstract val code: String
+
+    /** Imports to add because they are referenced by generated code. */
+    val imports: List<Candidate> by lazy { calculateImports() }
```
ajalt: Lazy properties are admittedly bit of a strange way to structure this. I could instead do something like `abstract fun run(): Pair<String, List<Candidate>>`, although since the generation is stateful, you wouldn't want to call `run` twice. I don't feel strongly either way.

src/main/kotlin/org/elm/ide/intentions/TyFunctionGenerator.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    protected open fun calculateImports(): List<Candidate> {
+        val visibleTypes = ModuleScope.getVisibleTypes(file).all
+                .mapNotNullTo(mutableSetOf()) { it.name?.let { n -> Ref(it.moduleName, n) } }
+        // Hack in List since GlobalScope.getVisibleTypes can't return it
+        val visibleModules = importedModules + "List"
+        return declarations
```
ajalt: Since we know we'll need an encoder for every type contained in the root ty, I assume that we need to import them all. That's not necessarily the case if we reuse an existing encoder, but this approach seemed simpler than tracking them manually.

src/main/kotlin/org/elm/ide/intentions/TyFunctionGenerator.kt
```diff
@@ -0,0 +1,117 @@
...
+                .filter {
+                    it.module != moduleName &&
+                            (it.toRef() in unionsToExpose || it.module !in visibleModules && it.toRef() !in visibleTypes)
+                }
+                .map {
```
ajalt: I could potentially make this `mapTo(mutableSetOf())` to prevent duplicate candidates, but the import merge takes care of that anyway, so I didn't bother.

src/main/kotlin/org/elm/lang/core/resolve/scope/GlobalScope.kt
```diff
@@ -88,7 +88,8 @@ class GlobalScope private constructor(val clientFile: ElmFile) {
         return rest
     }
 
-
+    // this function should also return List, but, unlike all the other global types, Elm doesn't
```
ajalt: Not having `type List = List` anywhere causes so many problems.

src/main/kotlin/org/elm/lang/core/types/Ty.kt
```diff
@@ -209,3 +210,43 @@ fun Ty.allVars(includeAlias: Boolean = false): Sequence<TyVar> = sequence<TyVar>
         alias?.parameters?.forEach { yieldAll(it.allVars(includeAlias)) }
     }
 }
+
+data class DeclarationInTy(val module: String, val name: String, val isUnion: Boolean)
```
ajalt: Not a fan of this data class, but it's nicer than returning a `Triple`

src/main/kotlin/org/elm/lang/core/resolve/scope/GlobalScope.kt
klazuka: I know. I asked Evan if he could fix it but got no response. I'll bring it up again.

src/main/kotlin/org/elm/ide/intentions/TyFunctionGenerator.kt
klazuka: There must be something wrong in my import merge code because I noticed that every time I run "create decoder" in the same file, it will add `required` to the exposed list. Resulting in situations like this:

```
import Json.Decode.Pipeline exposing (required, required, required)
```

klazuka: as a side note, the 'add import' stuff is pretty gnarly. I originally wrote it in Java for Kamil's plugin and then ported it to Kotlin. It also doesn't help that it's so hard for us to emit Elm code right now. I badly need to look into what it would take to do the formatter integration.

src/main/kotlin/org/elm/ide/intentions/MakeDecoderIntention.kt
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    private fun gen(ty: Ty): String = when (ty) {
+        is TyRecord -> generateRecord(ty)
+        is TyUnion -> generateUnion(ty)
+        is TyVar -> "(Debug.todo \"Can't generate decoders for type variables\")"
+        is TyTuple -> generateTuple(ty)
+        is TyUnit -> "(${qual("succeed")} ())"
+        is TyFunction, TyInProgressBinding, is MutableTyRecord, is TyUnknown -> {
+            "(Debug.todo \"Can't generate encoder for type ${ty.renderedText(false, false)}\")"
```
klazuka: s/encoder/decoder

src/main/kotlin/org/elm/ide/intentions/MakeDecoderIntention.kt
```diff
@@ -0,0 +1,183 @@
...
+        val func = GeneratedFunction(name = name, paramTy = ty, paramName = "", body = body, qualifier = qualifierFor(ty.toRef()))
+        funcsByTy[ty] = func
+        return name
+    }
+
+    private fun generateDict(ty: TyUnion): String {
+        existing(ty)?.let { return it }
+        val k = ty.parameters[0]
+        return if (k !is TyUnion || !k.isTyString) {
+            "(\\_ -> Debug.todo \"Can't generate encoder for Dict with non-String keys\")"
```
klazuka: s/encoder/decoder

src/main/kotlin/org/elm/ide/intentions/MakeDecoderIntention.kt
```diff
@@ -0,0 +1,183 @@
...
+            "Debug.todo \"Can't generate decoder for ${ty.name}\""
+        } else {
+            unionsToExpose += ty.toRef()
+            val variants = decl.variantInference().value
+            if (variants.size == 1 && variants.values.first().size == 1) {
+                "    ${qual("map")} ${variants.keys.first()} ${gen(variants.values.first().first())}"
+            } else {
+                val branches = variants.entries.joinToString("\n\n                ") { (variant, params) ->
+                    val expr = when {
+                        params.isEmpty() -> "Decode.succeed ${qual(ty.module, variant)}"
```
klazuka: I think `succeed` needs to be wrapped by `qual()`

src/main/kotlin/org/elm/ide/intentions/MakeDecoderIntention.kt
```diff
@@ -0,0 +1,183 @@
...
+        val k = ty.parameters[0]
+        return if (k !is TyUnion || !k.isTyString) {
+            "(\\_ -> Debug.todo \"Can't generate encoder for Dict with non-String keys\")"
+        } else {
+            "(${qual("dict")} ${gen(ty.parameters[1])})"
+        }
+    }
+
+    private fun generateTuple(ty: TyTuple): String {
+        val decoders = ty.types.mapIndexed { i, it -> "(Decode.index $i ${gen(it)})" }.joinToString(" ")
```
klazuka: `index` needs to be wrapped by `qual()`

src/main/kotlin/org/elm/ide/intentions/MakeEncoderIntention.kt
```diff
@@ -0,0 +1,173 @@
...
+        val param = rootFunc?.paramName ?: root.renderParam()
+        val body = rootFunc?.body ?: "    $genBody $param"
+
+        buildString {
+            append("\n$functionName $param =\n")
+            append(body)
+
+            for (f in funcsByTy.values) {
+                append("\n\n\n")
+                append("${f.name} : ${f.qualifier}${f.paramTy.renderedText(false, false)} -> Encode.Value\n")
```
klazuka: `Encode.Value` needs to be wrapped by `qual()`

src/main/kotlin/org/elm/lang/core/resolve/scope/GlobalScope.kt
klazuka: I created an issue for it https://github.com/elm/core/issues/1037

src/main/resources/intentionDescriptions/MakeDecoderIntention/after.elm.template
```diff
@@ -0,0 +1,3 @@
+decode : Decode.Value Int
+<spot>decode int =
+  Encode.int int</spot>
```
klazuka: I would make the example a little more compelling

```
import Json.Decode exposing (Decoder)

type alias Person =
    { name : String
    , age : Int
    }

personDecoder: Decoder Person
 <spot>personDecoder =
    Json.Decode.succeed Person
        |> required "name" Json.Decode.string
        |> required "age" Json.Decode.int</spot>
```

src/main/resources/intentionDescriptions/MakeEncoderIntention/after.elm.template
```diff
@@ -0,0 +1,3 @@
+encode : Int -> Encode.Value
```
klazuka: Just like the decode example, I would try to make it a little more exciting by showing how it works with a record.

src/main/kotlin/org/elm/ide/intentions/TyFunctionGenerator.kt
ajalt: Hm. That's a bug in the import merge. I am relying on it being idempotent, so I don't check existing imports when emitting code.

src/main/kotlin/org/elm/ide/intentions/MakeDecoderIntention.kt
```diff
@@ -0,0 +1,184 @@
...
+                append("${f.name} =\n")
+                append(f.body)
+            }
+        }
+    }
+
+    override fun calculateImports(): List<Candidate> {
+        return when {
+            usedPipeline -> super.calculateImports() + listOf(
+                    Candidate(
```
klazuka: Probably should rename `Candidate` since it is now being used in a different way.

src/main/kotlin/org/elm/ide/intentions/TyFunctionGenerator.kt
klazuka: Yeah, both sides have their tradeoffs.

I think I would prefer the `run` approach because reading input -> output would be nicer. It seems unlikely that anyone would ever re-run it. And you could add a flag to `TyFunctionGenerator` to keep track of this and throw when someone tries causing trouble.

But if that code comes out a lot worse, than this approach is totally fine with me.

src/main/kotlin/org/elm/lang/core/lookup/ElmLookup.kt
```diff
@@ -65,9 +66,28 @@ object ElmLookup {
...
 
+    /** Like [findByNameAndModule], but in the case of ambiguity, returns the first match located in the [file] */
+    inline fun <reified T : ElmNamedElement> findFirstByNameAndModule(
+            name: String,
+            module: String,
+            file: ElmFile
+    ): T? {
+        val candidates = findByNameAndModule<T>(name, module, file)
+        return when {
+            candidates.size < 2 -> candidates.firstOrNull()
```
klazuka: I know I wrote this code, and so I have only myself to blame, but the `< 2 ->` part is really gross to read. Would you mind changing this to an `if/else`?

src/main/kotlin/org/elm/ide/intentions/MakeDecoderIntention.kt
```diff
@@ -0,0 +1,184 @@
...
+                |
+                |                _ ->
+                |                    ${qual("fail")} ("unknown value for ${ty.renderedText(false, false)}: " ++ id)
+                |    in
+                |    ${qual("string")} |> ${qual("andThen")} get
+                """.trimMargin()
+            }
+        }
+
+        val name = "decode${ty.alias?.let { funcNames[it.toRef()] } ?: ty.name}"
```
klazuka: Can you change it so that generated, nested functions are named `fooDecoder` instead of `decodeFoo`? This is the style that elm-spa-example and the [elm-guide](https://guide.elm-lang.org/effects/json.html) uses.

src/main/kotlin/org/elm/ide/intentions/MakeDecoderIntention.kt
```diff
@@ -0,0 +1,184 @@
...
+        else -> generateUnionFunc(ty)
+    }
+
+    private fun generateRecord(ty: TyRecord): String {
+        val alias = ty.alias ?: return "(Debug.todo \"Cannot decode records without aliases\")"
+        val cached = existing(ty) ?: funcsByTy[ty]?.name
+        if (cached != null) return cached
+
+        val qualifier = qualifierFor(alias.toRef())
+        val name = "decode${funcNames[alias.toRef()]}"
```
klazuka: see comment about nested function name in `generateUnionFunc`

src/main/kotlin/org/elm/ide/intentions/MakeDecoderIntention.kt
```diff
@@ -0,0 +1,187 @@
...
+        funcsByTy.remove(root)
+        val body = rootFunc?.body ?: "    $genBody"
+
+        return buildString {
+            append("\n$functionName =\n")
+            append(body)
+
+            for (f in funcsByTy.values) {
+                append("\n\n\n")
+                append("-- TODO: generated code\n")
```
ajalt: Is this good, or should this message be more descriptive to make sure people know that the code probably need editing?

klazuka: how about `-- TODO: double-check generated code`?

PR #458, opened 2019-08-06
Remove inline class usage

ajalt:
As of kotlin 1.3.41, using inline classes causes a compiler warning about using them in production code due to api instability. This message pops up over your build or debug window every time you run a test in IntelliJ, which is pretty annoying during development.

This PR swaps out the inline class for extension methods, which are still terse, but don't use unstable language features. Feel free to close this PR if you'd prefer to keep the current implementation.


klazuka:
I was starting to have reservations myself, especially since, due to the plugin nature of the code, we may run in an environment which we were not compiled against.

PR #459, opened 2019-08-06
Support references to record fields in annotated functions

ajalt:
This allows features such as Go To Definition, Rename, Quick Documentation, and Find Usages to work on record fields. This works for records defined in either type aliases or type annotations. It should work for all field names inside of a top-level annotated function, including within nested functions. This includes field access expressions, field accessor functions, and record value expressions that get assigned to a type alias or union.

Only references to unions and type aliases will resolve. Field access of record literals, either in the same function or in another unannotated function, will not resolve. Field access on the parameters of unannotated top-level functions will not resolve.

As part of this change, I removed the pin from the field access expression bnf rule so that the name element is always present. This causes error messages to be slightly more cryptic in some cases, but it also fixes the incorrect parsing of expressions like `a .b`.

Capture

---

Capture2

---

Capture3


src/main/kotlin/org/elm/lang/core/types/Ty.kt
```diff
@@ -59,11 +59,14 @@ data class TyTuple(val types: List<Ty>, override val alias: AliasInfo? = null) :
...
  * @property alias The alias for this record, if there is one. Used for rendering and tracking record constructors
+ * @property fieldReferences A map of field name to the psi element that defines them; used for
+ *   reference resolve, does not affect equality
  */
 data class TyRecord(
         val fields: Map<String, Ty>,
         val baseTy: Ty? = null,
-        override val alias: AliasInfo? = null
+        override val alias: AliasInfo? = null,
+        val fieldReferences: RecordFieldReferenceTable = RecordFieldReferenceTable()
```
ajalt: I really don't like this hack of having a mutable collection of PsiElements tacked on to a Ty, but it is the best solution I could come up with since it works well with all the unification code. I tried to mitigate the potential downsides by freezing this table when a ty is cached to prevent modification of cached values.

src/main/kotlin/org/elm/lang/core/types/TypeReplacement.kt
```diff
@@ -58,7 +59,29 @@ class TypeReplacement(
...
-            return TypeReplacement(emptyMap(), freshen = false, varsToRemainRigid = emptyList()).replace(ty)
+            return TypeReplacement(emptyMap(), freshen = false, varsToRemainRigid = emptyList(), freeze = false).replace(ty)
+        }
+
+        /**
+         * Freeze any fields references present in records in a [ty].
+         *
+         * Use this to prevent cached values from being altered.
+         */
+        fun freeze(ty: Ty): Ty {
```
ajalt: I don't know if it great that we can reuse type replacement or so many things, or problematic that we do.

src/test/resources/org/elm/lang/core/parser/fixtures/partial/FieldAccessors.txt
```diff
@@ -8,57 +8,77 @@ Elm File
...
-    ElmFieldAccessExpr(FIELD_ACCESS_EXPR)
-      ElmValueExpr(VALUE_EXPR)
-        ElmValueQID(VALUE_QID)
-          PsiElement(UPPER_CASE_IDENTIFIER)('Foo')
-          PsiElement(DOT)('.')
-          PsiElement(LOWER_CASE_IDENTIFIER)('bar')
-      PsiElement(DOT)('.')
-      PsiErrorElement:<expr> or LOWER_CASE_IDENTIFIER expected
-        <empty list>
+    ElmParenthesizedExpr(PARENTHESIZED_EXPR)
```
ajalt: GitHub isn't diffing this well, but pretty much only the error elements changed here.

src/main/kotlin/org/elm/lang/core/psi/elements/ElmFieldAccessorFunctionExpr.kt
```diff
@@ -2,18 +2,33 @@ package org.elm.lang.core.psi.elements
...
+
+    override val referenceNameElement: PsiElement
+        get() = identifier
+
+    override val referenceName: String
+        get() = identifier.text
+
+    override fun getReference(): ElmReference {
+        return object : RecordFieldReference<ElmFieldAccessorFunctionExpr>(this@ElmFieldAccessorFunctionExpr) {
+            override fun getTy(): Ty? = (element.findTy() as? TyFunction)?.parameters?.singleOrNull()
```
klazuka: Doesn't type inference rely on the reference/resolve system? Is there any danger of an infinite loop here?

src/main/kotlin/org/elm/lang/core/types/Ty.kt
```diff
@@ -72,9 +75,23 @@ data class TyRecord(
...
+            val prefix = if (fieldReferences.isEmpty()) "" else "+"
+            val params = if (it.parameters.isEmpty()) "" else " ${it.parameters.joinToString(" ")}"
+            "{$prefix${it.name}$params}"
         } ?: baseTy?.let { "{$baseTy | $f}" } ?: "{$f}"
     }
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other !is TyRecord) return false
+        return fields == other.fields && baseTy == other.baseTy
```
klazuka: I'm sure you know what you're doing, but just to double-check, should `alias` be included in equality/hash?

src/main/kotlin/org/elm/lang/core/types/TypeReplacement.kt
```diff
@@ -44,9 +45,9 @@ class TypeReplacement(
          * `f : (a -> b) -> a -> b`, every time you call the `(a -> b)` parameter within the body of
          * `f`, the ty of `a` and `b` have to be the same.
          */
-        fun freshenVars(ty: Ty): Ty {
-            if (ty.allVars(true).none()) return ty
-            return TypeReplacement(emptyMap(), freshen = true, varsToRemainRigid = null).replace(ty)
+        fun freshenVars(ty: Ty, freeze: Boolean = false): Ty {
+            if (!tyWouldChange(ty, freeze)) return ty
+            return TypeReplacement(emptyMap(), freshen = true, varsToRemainRigid = null, freeze = false).replace(ty)
```
klazuka: This is probably ok, but just to double-check, is it ok that the `freeze` param is always `false` here?

src/main/kotlin/org/elm/lang/core/psi/elements/ElmFieldAccessorFunctionExpr.kt
ajalt: It does use the resolve system, but it never resolves the new reference types added in the PR, so there's no change of a loop. I'll add a comment warning future maintainers about that.

src/main/kotlin/org/elm/lang/core/types/TypeReplacement.kt
ajalt: Nope, that's a copy-paste error. Good catch!

src/main/kotlin/org/elm/lang/core/types/Ty.kt
ajalt: It doesn't actually affect anything, since only TyVars are used as map keys, and I never compare tys by equality. But it's probably better to add it for consistency with other data class tys.

PR #463, opened 2019-08-07
Add quick fix to rename unused patterns to _

ajalt:
The kotlin plugin provides a similar fix. This is more useful for case patterns than it is for function parameters, but there's no harm in providing it for both.

Capture2

Capture

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/psi/ElmPsiFactory.kt
+++ b/src/main/kotlin/org/elm/lang/core/psi/ElmPsiFactory.kt
@@ -105,22 +105,23 @@ class ElmPsiFactory(private val project: Project) {
                     ?: error("Invalid upper-case identifier: `$text`")
 
     fun createUpperCaseQID(text: String): ElmUpperCaseQID =
-            createFromText<ElmModuleDeclaration>("module $text exposing (..)")
-                    ?.upperCaseQID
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/elm/lang/core/psi/ElmPsiFactory.kt
```diff
@@ -105,22 +105,23 @@ class ElmPsiFactory(private val project: Project) {
...
-            createFromText<ElmModuleDeclaration>("module $text exposing (..)")
-                    ?.upperCaseQID
+            createFromText("module $text exposing (..)")
                     ?: error("Invalid upper-case QID: `$text`")
 
     fun createValueQID(text: String): ElmValueQID =
-            createFromText<ElmValueDeclaration>("f = $text")
-                    ?.expression
-                    ?.childOfType()
+            createFromText("f = $text")
```
ajalt: There are some unrelated cleanups in this file. I can revert them if you'd like.

klazuka: sorry, I touched this in my PR. Can you rebase?

ajalt: done

PR #464, opened 2019-08-08
Reduce stubs boilerplate

klazuka:
Last night I started looking at putting type annotations into the stub system, but it was going to require a ton of boilerplate. Luckily intellij-rust has already dealt with the problem. This should make it a little bit easier to do.

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/psi/PsiElement.kt
+++ b/src/main/kotlin/org/elm/lang/core/psi/PsiElement.kt
@@ -26,13 +26,14 @@ SOFTWARE.
 
 package org.elm.lang.core.psi
 
-import com.intellij.psi.PsiComment
-import com.intellij.psi.PsiElement
-import com.intellij.psi.PsiErrorElement
-import com.intellij.psi.PsiWhiteSpace
+import com.intellij.extapi.psi.StubBasedPsiElementBase
+import com.intellij.psi.*
+import com.intellij.psi.impl.source.PsiFileImpl
+import com.intellij.psi.stubs.StubElement
 import com.intellij.psi.tree.IElementType
 import com.intellij.psi.util.PsiTreeUtil
 import com.intellij.psi.util.PsiUtilCore
+import com.intellij.util.SmartList
 import org.elm.lang.core.lexer.ElmLayoutLexer
 import org.elm.lang.core.psi.ElmTypes.VIRTUAL_END_DECL
 import org.elm.lang.core.psi.elements.ElmValueDeclaration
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/elm/lang/core/psi/PsiElement.kt
```diff
@@ -72,12 +73,106 @@ fun <T : PsiElement> PsiElement.parentOfType(vararg classes: KClass<out T>): T?
 inline fun <reified T : PsiElement> PsiElement.contextOfType(strict: Boolean = true): T? =
         PsiTreeUtil.getContextOfType(this, T::class.java, strict)
 
-inline fun <reified T : PsiElement> PsiElement.childOfType(strict: Boolean = true): T? =
-        PsiTreeUtil.findChildOfType(this, T::class.java, strict)
+inline fun <reified T : PsiElement> PsiElement.descendantOfType(): T? =
```
ajalt: Love this rename; I've had to look through the source code multiple times to remember whether this was only direct children or not.

src/main/kotlin/org/elm/lang/core/psi/PsiElement.kt
```diff
@@ -72,12 +73,106 @@ fun <T : PsiElement> PsiElement.parentOfType(vararg classes: KClass<out T>): T?
...
+/** Returns direct children stubs (or AST, if necessary) of the specified type */
+inline fun <reified T : PsiElement> PsiElement.stubDirectChildrenOfType(): List<T> {
+    return if (this is PsiFileImpl) {
+        stub?.childrenStubs?.mapNotNull { it.psi as? T } ?: return directChildrenOfType()
+    } else {
+        PsiTreeUtil.getStubChildrenOfTypeAsList(this, T::class.java)
+    }
+}
+
+inline fun <reified T : PsiElement> PsiElement.stubDescendantsOfTypeStrict(): Collection<T> =
```
ajalt: Maybe use a `strict` parameter rather than separate functions for the strict case? It's fine either way.

klazuka: I copied this straight from intellij-rust. I don't have a preference either, at least not yet.

src/main/kotlin/org/elm/lang/core/psi/PsiElement.kt
klazuka: I know, I can't believe that they have built so much code on such poor naming conventions. There were undoubtedly a lot of bugs along the way.

PR #465, opened 2019-08-08
Add intention to qualify unresolved references

### LINKED ISSUE #419: Add intention to add qualifier to fix unresolved ref
When a bare reference is unresolved, but could be resolved if qualified with an imported module name or alias, we should provide an intention to add the qualifier.

ajalt:
This intention is applicable to unresolved references to names defined in imported modules where the name isn't exposed in the current module. It will add the qualifier prefix necessary to make the reference resolve.

Capture

---

Capture2

---

Capture3

Fixes #419

### DIFF
--- a/src/main/kotlin/org/elm/ide/intentions/AddImportIntention.kt
+++ b/src/main/kotlin/org/elm/ide/intentions/AddImportIntention.kt
@@ -31,7 +31,7 @@ interface ImportPickerUI {
 private var MOCK: ImportPickerUI? = null
 
 @TestOnly
-fun withMockUI(mockUi: ImportPickerUI, action: () -> Unit) {
+fun withMockImportPickerUI(mockUi: ImportPickerUI, action: () -> Unit) {
     MOCK = mockUi
     try {
         action()
@@ -81,16 +81,16 @@ class AddImportIntention : ElmAtCaretIntentionActionBase<AddImportIntention.Cont
                 // but if it's an alias-based import, we should show the user some UI so that
                 // they know what they're getting into. See https://github.com/klazuka/intellij-elm/issues/309
                 when {
-                    candidate.moduleAlias != null -> promptToSelectCandidate(context, file)
+                    candidate.moduleAlias != null -> promptToSelectCandidate(project, context, file)
                     else -> addImport(candidate, file, context.isQualified)
                 }
             }
-            else -> promptToSelectCandidate(context, file)
+            else -> promptToSelectCandidate(project, context, file)
         }
     }
 
 
-    private fun promptToSelectCandidate(context: Context, file: ElmFile) {
+    private fun promptToSelectCandidate(project: Project, context: Context, file: ElmFile) {
         require(context.candidates.isNotEmpty())
 
         // Put exact matches (i.e. those with `moduleAlias == null`) at the top of the list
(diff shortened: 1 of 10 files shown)


src/main/kotlin/org/elm/ide/intentions/AddImportIntention.kt
```diff
@@ -81,16 +81,16 @@ class AddImportIntention : ElmAtCaretIntentionActionBase<AddImportIntention.Cont
...
                 }
             }
-            else -> promptToSelectCandidate(context, file)
+            else -> promptToSelectCandidate(project, context, file)
         }
     }
 
 
-    private fun promptToSelectCandidate(context: Context, file: ElmFile) {
+    private fun promptToSelectCandidate(project: Project, context: Context, file: ElmFile) {
```
klazuka: Why add `project` parameter? It can be derived from the `file` parameter (or alternatively put it in the `Context`)

src/main/kotlin/org/elm/ide/intentions/AddQualifierIntention.kt
```diff
@@ -0,0 +1,126 @@
...
+            }
+        }
+    }
+
+    private fun makeContext(file: ElmFile, qid: ElmQID, ref: ElmReferenceElement, names: VisibleNames): Context? {
+        val name = ref.referenceName
+        val candidates = (names.global + names.imported)
+                .filter { it.name == name }
+                .mapNotNull { ModuleScope.getQualifierForName(file, it.moduleName, name) }
+                .filter { it.isNotEmpty() }
```
klazuka: nbd but when I read this I assumed that `it` was a collection. If you change this to `filter { it != "" }` it would be easier to read.

src/main/kotlin/org/elm/ide/intentions/AddQualifierIntention.kt
```diff
@@ -0,0 +1,126 @@
...
+
+    override fun invoke(project: Project, editor: Editor, context: Context) {
+        when (context.candidates.size) {
+            0 -> error("should not happen: must be at least one candidate")
+            1 -> addQualifier(project, context, context.candidates.first())
+            else -> promptToSelectCandidate(project, editor, context)
+        }
+    }
+
+    private fun promptToSelectCandidate(project: Project, editor: Editor, context: Context) {
```
klazuka: this method and `addQualifier` should be swapped so that it reads top-down the same way that you read the `when` in `invoke`

src/main/kotlin/org/elm/ide/intentions/AddQualifierIntention.kt
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        val qid = element.parentOfType<ElmQID>() ?: return null
+        if (qid.isQualified) return null
+
+        val typeRef = qid.parentOfType<ElmTypeRef>()
+        if (typeRef != null) {
+            return makeContext(file, qid, typeRef, ModuleScope.getReferencableTypes(file))
+        }
+
```
klazuka: I noticed that the intention is not available for `BadStatus` in the following:

```elm
import Http

decodeErrors : Http.Error -> List String
decodeErrors error =
    case error of
        BadStatus response ->
            [ "bad" ]

        err ->
            [ "Server error" ]
```

The problem is that the `qid`'s parent in this case is `ElmUnionPattern` not `ElmValueExpr`.

src/main/kotlin/org/elm/ide/intentions/AddImportIntention.kt
ajalt: Just because we already have it, so there's no need to derive. It's less code overall. I can revert this if you want.

klazuka: I guess my preference is always less-code-to-read vs less-code-to-execute. But I don't care either way here. I was just curious.

Issue #428, opened 2019-07-09
Exception in plugin Elm (3.1.2)


emthornber:
com.intellij.openapi.extensions.impl.PicoPluginExtensionInitializationException: org.frawa.elmtest.run.ElmTestRunConfigurationProducer has unsatisfied dependency: class org.frawa.elmtest.run.ElmTestConfigurationFactory among unsatisfiable dependencies: [[class org.frawa.elmtest.run.ElmTestConfigurationFactory], [class org.frawa.elmtest.run.ElmTestRunConfigurationType]] where DefaultPicoContainer (root) was the leaf container being asked for dependencies.

Don't know what triggered the exception.  It's the second time today so I'll keep an eye out.

klazuka:
Maybe you had installed the [intellij-elm-test-runner](https://github.com/frawa/intellij-elm-test-runner)  plugin before that code got merged directly into this plugin? If so, it could be causing a conflict and you should uninstall it.

emthornber:
Keith,

Ah - hadn’t realised it was Either/Or for those plugins.  I will uninstall
and report back.

AAMOI - does the plugin recognise that there is already an index.html and
automatically output a .js file when compiling ?

Thanks
—
MarkT

On 9 July 2019 at 15:17:08, Keith Lazuka (notifications@github.com) wrote:

Maybe you had installed the intellij-elm-test-runner
<https://github.com/frawa/intellij-elm-test-runner> plugin before that code
got merged directly into this plugin? If so, it could be causing a conflict
and you should uninstall it.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/klazuka/intellij-elm/issues/428?email_source=notifications&email_token=AACLEF4NHCSPQYFAXQSASNLP6SMWHA5CNFSM4H7FOY4KYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGODZQNAEQ#issuecomment-509661202>,
or mute the thread

klazuka:
No, the current integration with the Elm compiler is limited: it only runs the compiler in order to detect compile-time errors. Most compile-time errors should be shown directly within IntelliJ automatically because of our parser, reference resolver and type checker, but there may be some things that slip through (especially in un-opened files). So we provide the `elm` binary as the final safety-net. The generated code goes to `/dev/null`.

In the future I would like to make this work so that it can generate and run code.

[...]

Issue #387, opened 2019-04-09
Slow rename in huge files


klazuka:
Renaming a function is very slow in big files (1,000+ lines). Simon sent me a CPU trace from his machine and it showed excessive CPU thrashing in the parser:

```
1. Atom
2. RecordExpr
3. RecordInner
4. BinOpExprUpper
5. LetInTail
```

He reports that inline rename is lagging in big files by ~1 second every keystroke. I tested it on the remedy editor in PowerLift (1,800 lines) and it was slow, but not quite that slow (maybe 100-500ms).

klazuka:
One mitigation would be to disable inplace rename in huge files. But it would be better to just fix the parser (assuming that it's related to excessive backtracking).

PR #467, opened 2019-08-11
Use stubs during type inference

klazuka:
I noticed that when opening a file in IntelliJ shortly after opening a project, it was also parsing nearly 100 other Elm files imported from that file (either directly or indirectly). This was happening because the type inference code depended on data which was not present in the stub-backed Psi tree. I'm actually kinda surprised that there wasn't a more noticeable performance impact, but I guess since it was being done on a background thread, and since the results were cached, it wasn't too big a deal.

In any case, the right thing to do is to make sure that type inference has everything it needs in the stub tree, and that's what this PR does. This primarily means making type definitions and type annotations stub-backed, but there were also a lot of smaller things that I had to make into stubs to make things work.

The end result is that now when you open `Page.Remedy.Editor` from the PowerLift project, it only parses a single file instead of 95 files in the old implementation. It was hard for me to measure the performance impact, but it looks like total analysis time for that file went from 5 seconds to 4 seconds. Not quite as big of an improvement as I expected, but I suspect that there are other performance problems which dominate (#387 and #428 ).


klazuka:
@ajalt this is going to be pretty hard to review. I tried to do it baby steps where I made sure that all the tests passed with each refactoring/commit. The core of it is in `StubImplementations.kt`, and the tests are in `TypeInferenceStubAccessTest`

PR #468, opened 2019-08-12
Convert UnresolvedReferenceAnnotator to an intention

ajalt:
Intentions have extra functionality like the ability to be run on the entire project, disabled per-file, or suppressed with a comment. The new intention otherwise behaves the same as the annotator. 

As part of this change, I had to convert the intentions surfaced by the annotator into quick fixes. I also changed the unresolved symbol highlight to the standard color.


src/main/kotlin/org/elm/ide/inspections/ElmUnresolvedReferenceInspection.kt
```diff
@@ -47,14 +35,11 @@ class ElmUnresolvedReferenceAnnotator : Annotator {
...
-        // Give each handler a chance to deal with the unresolved ref before falling back on an error
-        outerLoop@
         for (ref in refs.filter { it.resolve() == null }) {
-            for (handler in handlers) {
-                if (handler(ref, element, holder)) {
-                    continue@outerLoop
-                }
-            }
+            // Give each handler a chance to deal with the unresolved ref before falling back on an error
+            if (handleTypeAnnotation(ref, element, holder)) continue
```
klazuka: this is a lot better. I think I had been in Spring-land too long when I thought of the handler thing 😄

Birowsky:
Sweet!

Birowsky:
What does 'running on entire project' actually mean? What was not running on entire project so far?

klazuka:
before the annotator would only run on files that you had open in the editor. But now you can do "run inspection by name" -> "unresolved reference", and it will check all files in your project (or whatever scope you limit it to).

PR #477, opened 2019-08-20
Fix case where record field reference wouldn't resolve

### LINKED ISSUE #475: Record field renaming not works in some specific cases (really deep)
here i rename it:
https://github.com/justgook/platformer/blob/develop/lib/elm-logic-templates/src/Logic/Template/Game/ShootEmUp.elm#L138

here it is not renamed:
https://github.com/justgook/platformer/blob/develop/lib/elm-logic-templates/src/Logic/Template/Game/ShootEmUp.elm#L561

ajalt:
To my surprise, it turns out that if you try _really_ hard, you can in fact assign an extension record to a concrete record before the concrete record is unified with a type annotation. This commit now tracks replacements for that case.

Note that this didn't affect inference correctness, it just prevented us from propagating field references.

Fixes #475

### DIFF
--- a/src/test/kotlin/org/elm/lang/core/resolve/ElmRecordFieldResolveTest.kt
+++ b/src/test/kotlin/org/elm/lang/core/resolve/ElmRecordFieldResolveTest.kt
@@ -265,4 +265,24 @@ main =
         --^                               
     ]
 """)
+
+    fun `test nested extension aliases with funcion in type variable passed through another variable via forward pipeline`() = checkByCode(
+            """
+infix left  0 (|>) = apR
+apR : a -> (a -> b) -> b
+apR x f = f x
+
+type alias R = { field : () }
+                  --X
+type alias Outer r = { r : Type (r -> r) }
+type Type a = Type a
+
+foo : Outer r -> Outer r
+foo r = r
+
+main : Outer R
+main =
+    { r = Type (\r -> { r | field = () }) } |> foo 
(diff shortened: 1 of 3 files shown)


src/test/kotlin/org/elm/lang/core/resolve/ElmRecordFieldResolveTest.kt
```diff
@@ -265,4 +265,24 @@ main =
         --^                               
     ]
 """)
+
+    fun `test nested extension aliases with funcion in type variable passed through another variable via forward pipeline`() = checkByCode(
```
ajalt: ¯\\\_(ツ)_/¯

src/test/kotlin/org/elm/lang/core/resolve/ElmRecordFieldResolveTest.kt
```diff
@@ -265,4 +265,24 @@ main =
...
+                  --X
+type alias Outer r = { r : Type (r -> r) }
+type Type a = Type a
+
+foo : Outer r -> Outer r
+foo r = r
+
+main : Outer R
+main =
+    { r = Type (\r -> { r | field = () }) } |> foo
```
ajalt: It seems like there should be a smaller test case for this, but this is as much as I could minify it. For example, the problem isn't present if you use `<|` or a regular function call on this line.

PR #478, opened 2019-08-20
Resolve record pattern references

### LINKED ISSUE #476: Plugin reports unused record fields in case of record destructuring
Hey there!

First of all let me express my big gratitude for this plug-in as it is helping enormously and brings my developer experience to great heights. Existence of this plug-in, its rich set of features and overall great reliability often make it easy to do changes in several files without hitting Cmd+S single time and thus not seeing any compiler error which is a personal point of satisfaction. :-)

However, after updating to 3.4.0 I have immediately discovered that it started marking some of my record fields as unused for the sole reason of using these records in a destructured manner only. This brings reliability down significantly as I know I can't rely on usage search anymore and have to resort to full-text search again. 

Please find SSCCE in [Ellie](https://ellie-app.com/6r6rm5ftdRda1) or below. Both fields in `Flags` are marked as unused despite on the type annotations.

**Screenshot:**
Screenshot 2019-08-20 at 13 04 44

**Source:** (also in [Ellie](https://ellie-app.com/6r6rm5ftdRda1))
```
module Sscce exposing (main)

import Browser exposing (Document)


type alias Flags =
    { flag1 : Int
    , flag2 : String
    }


type alias Model =
    ( Int, String )


init : Flags -> ( Model, Cmd () )
init ({ flag1, flag2 } as flags) =
    let
        _ =
            Debug.log "TEST" ( test1 flags, test2 flags )
    in
    ( ( flag1, flag2 )
    , Cmd.none
    )


test1 : Flags -> String
test1 { flag1 } =
    String.fromInt flag1


test2 : Flags -> String
test2 { flag2 } =
    flag2


main : Program Flags Model ()
main =
    Browser.document
        { init = init
        , update = \() model -> ( model, Cmd.none )
        , view = always (Document "SSCCE" [])
        , subscriptions = always Sub.none
        }
```

ajalt:
This took two changes:

- Make ElmLowerPattern a reference element. It was already a named element; now it's both.
- Make LexicalValueReferences pass through parameters that are also references

The second change is necessary because otherwise find usages / renaming a record field would not find ElmValueExpr elements, since they would only resolve to the parameter rather than the field. Unfortunately, that change means that parameters that resolve to another element will always be marked as unused. I worked around that by ignoring those elements in the unused symbol inspection. It also means that we can't generate documentation for record pattern parameters that reference records directly inside type annotations (it still works for references to aliased records).

Given the imperfect nature of this fix, you may want to reject this PR. It seems that this feature is more useful than the ones it breaks, but that's up to you.

Fixes #476

### DIFF
--- a/src/main/kotlin/org/elm/ide/inspections/ElmUnusedSymbolInspection.kt
+++ b/src/main/kotlin/org/elm/ide/inspections/ElmUnusedSymbolInspection.kt
@@ -21,7 +21,10 @@ class ElmUnusedSymbolInspection : ElmLocalInspection() {
         val name = element.name
 
         // ignore certain kinds of declarations which we don't want to inspect
-        if (element is ElmTypeAliasDeclaration || element is ElmTypeDeclaration || element is ElmTypeVariable) {
+        if (element is ElmTypeAliasDeclaration ||
+                element is ElmTypeDeclaration ||
+                element is ElmTypeVariable ||
+                element is ElmLowerPattern && element.parent is ElmRecordPattern) {
             // TODO revisit: implementation for types is a little tricky since
             //      type annotations are optional; punting for now
             return
--- a/src/main/kotlin/org/elm/lang/core/psi/elements/ElmLowerPattern.kt
+++ b/src/main/kotlin/org/elm/lang/core/psi/elements/ElmLowerPattern.kt
@@ -6,6 +6,9 @@ import com.intellij.psi.search.LocalSearchScope
 import com.intellij.psi.search.SearchScope
 import com.intellij.psi.util.PsiTreeUtil
 import org.elm.lang.core.psi.*
+import org.elm.lang.core.resolve.ElmReferenceElement
+import org.elm.lang.core.resolve.reference.ElmReference
+import org.elm.lang.core.resolve.reference.RecordFieldReference
 
 
 /**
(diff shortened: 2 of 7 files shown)


klazuka:
I'm not strictly opposed to making it both a reference and a named element (intellij-rust [does it](https://github.com/intellij-rust/intellij-rust/blob/5cfa05c8294d1fe8d19cff9e78401e45da648e8b/src/main/kotlin/org/rust/lang/core/psi/ext/RsPatBinding.kt#L56)), but it seems like we could avoid the problem entirely by changing the parse rule for `RecordPattern` so that it doesn't use `LowerPattern` but instead uses some new reference element that holds a lower-case identifier. The downside is that calling code (like the type inference algorithm) would need to handle yet-another type of psi element.

I'm not sure if the increased simplicity would be worth the extra code. What do you think?

ajalt:
We also already have an element that's [both a named element and a reference](https://github.com/klazuka/intellij-elm/blob/master/src/main/kotlin/org/elm/lang/core/psi/elements/ElmTypeVariable.kt). With this PR, value references will resolve to the field pattern parameter itself if there's no type annotation. If we used a new element that wasn't a named element, then references to record patterns in functions without annotations would not resolve, and renaming those parameters would not rename the references etc.

klazuka:
Good point!

PR #484, opened 2019-08-22
Fix inference errors in field patterns of unannotated records

### LINKED ISSUE #480: intellij type checker reports error in valid code
this code compiles fine - see https://ellie-app.com/6rzCpH8bczCa1 - but has an error in intellij

and here is a picture of a different issue for which I do not have an SSCCSSE
https://files.slack.com/files-pri/T0CJ5UNHK-FMCN83ZJM/image.png

ajalt:
- Check for MutableTyRecord when binding field patterns
- Keep records mutable when replacing function calls

Fixes #480

### DIFF
--- a/src/test/kotlin/org/elm/ide/inspections/inference/TypeInferenceInspectionTest1.kt
+++ b/src/test/kotlin/org/elm/ide/inspections/inference/TypeInferenceInspectionTest1.kt
@@ -290,6 +290,15 @@ main : ()
 main = foo .x
 """)
 
+    fun `test field accessor as argument to type var in pipeline`() = checkByText("""
+foo : (a -> ()) -> a -> a
+foo _ a = a
+
+main : ()
+main =
+    <error descr="Type mismatch.Required: ()Found: { a | f2 : (), f1 : () }">{ f1 = (), f2 = () } |> foo .f2</error>
+""")
+
     fun `test correct value type from parametric record alias`() = checkByText("""
 type alias A a = {x: a, y: ()}
 main : A Float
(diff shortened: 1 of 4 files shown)


src/test/kotlin/org/elm/ide/inspections/inference/TypeInferenceInspectionTest1.kt
```diff
@@ -290,6 +290,15 @@ main : ()
 main = foo .x
 """)
 
+    fun `test field accessor as argument to type var in pipeline`() = checkByText("""
+foo : (a -> ()) -> a -> a
+foo _ a = a
+
+main : ()
+main =
+    <error descr="Type mismatch.Required: ()Found: { a | f2 : (), f1 : () }">{ f1 = (), f2 = () } |> foo .f2</error>
```
ajalt: This isn't entirely correct: the actual type should be `{ f2 : (), f1 : () }` i.e. not a subset. But this won't cause false positives, and will only cause false negatives in very contrived code. It doesn't seem worth fixing right now.

klazuka:
Looks like the new test may have failed on Azure DevOps:
```
 expected:<...equired: ()Found: { [a | f2 : (), f1] : () }">{ f1 = (), ...> but was:<...equired: ()Found: { [f1 : (), f2] : () }">{ f1 = (), ...>
```

Maybe there is some kind of unexpected thing going on here? it looks like the actual output was what you said it should be in the comment.

ajalt:
I don't know why that test was failing on azure but not on circle or locally. I just removed it, since it was testing for technically incorrect output anyway, and the changes in this PR are covered by other tests.

PR #500, opened 2019-09-03
Use type annotations when inferring recursive functions.

### LINKED ISSUE #499: Misses a typecheck issue
Hi, guys!
First of all, thanks for the plugin it makes my life much better :)

I've stumbled upon a small type inference issue, this code appears normal to IntelliJ but in reality it doesn't typecheck:

https://ellie-app.com/6y2DnFD4zVza1

```elm
import Bytes.Decode as D


type Tree a
    = Node (List (Tree a))
    | Leaf a


sequenceTree : Tree (D.Decoder a) -> D.Decoder (Tree a)
sequenceTree tree =
    case tree of
        Node trees ->
            Node >> D.succeed <| List.map sequenceTree trees

        Leaf decoder ->
            D.map Leaf decoder

```

image

ajalt:
Currently, we are returning `TyUnknown` for all referenced recursive value declarations to avoid infinite loops. This is overly cautious and results in some false negatives.

This commit uses type annotations for recursive functions when available, since there's no risk of a loop in that case. We still return `TyUnknown` when no annotation is present.

Fixes #499

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
+++ b/src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
@@ -117,8 +117,7 @@ private class InferenceScope(
         // If the assignee has any syntax errors, we don't run inference on it. Trying to resolve
         // references in bodies when the parameters contain syntax errors can result in infinite
         // recursion and surprising bugs.
-        // We don't currently infer recursive functions in order to avoid infinite loops in the inference.
-        if (checkRecursion(declaration) || assignee == null || assignee.hasErrors) {
+        if (assignee == null || assignee.hasErrors) {
             return toTopLevelResult(TyUnknown())
         }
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
```diff
@@ -117,8 +117,7 @@ private class InferenceScope(
         // If the assignee has any syntax errors, we don't run inference on it. Trying to resolve
         // references in bodies when the parameters contain syntax errors can result in infinite
         // recursion and surprising bugs.
-        // We don't currently infer recursive functions in order to avoid infinite loops in the inference.
-        if (checkRecursion(declaration) || assignee == null || assignee.hasErrors) {
+        if (assignee == null || assignee.hasErrors) {
```
ajalt: This check was superfluous with the check  in `inferReferencedValueDeclaration`.

PR #505, opened 2019-09-04
Infer types of references to record pattern parameters without aliases

ajalt:
Starting with PR #478, we always inferred references to record field pattern parameters of functions annotated with raw record types as `TyUnknown`. This PR fixes that regression.

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/resolve/reference/LexicalValueReference.kt
+++ b/src/main/kotlin/org/elm/lang/core/resolve/reference/LexicalValueReference.kt
@@ -14,8 +14,17 @@ class LexicalValueReference(element: ElmReferenceElement)
             emptyArray()
 
     override fun resolveInner(): ElmNamedElement? {
-        val resolved = getCandidates().find { it.name == element.referenceName }
+        val resolved = resolveShallow()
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/elm/lang/core/resolve/reference/LexicalValueReference.kt
```diff
@@ -14,10 +14,19 @@ class LexicalValueReference(element: ElmReferenceElement)
...
+        val resolved = resolveShallow()
         return (resolved as? ElmReferenceElement)?.reference?.resolve() ?: resolved
     }
 
+    /**
+     * References of this type normally recursively resolve references to elements that are
+     * themselves references (e.g. usage of a name in a record destructuring pattern). This function
+     * will only resolve the first level of reference.
+     */
+    fun resolveShallow(): ElmNamedElement? {
```
ajalt: Is this too much of a hack? The recursive resolution is already pretty hacky.

klazuka: It's not ideal, but I think this is the world we live in. I seem to remember similar things in intellij-rust.

PR #506, opened 2019-09-05
First pass on optimizing parser for deeply nested expressions

### LINKED ISSUE #387: Slow rename in huge files

### LINKED ISSUE #421: Major performance degradation parsing deeply nested functions
Medium to large edits in files often causes the plugin to freeze IntelliJ for a minute or so. I guess because it's trying to figure out errors, types or formatting. It then refreezes on every key press until the errors are fixed. 

I really would like a feature that either interrupts the computation after half a second or so, or do the computation in such a way that you can continue writing. 

I could maybe post an example, but it would be quite big. Any disruptive edit in a large project ought to reproduce the issue.

klazuka:
Fixes #387
Fixes #421

I found 2 problems:
1. backtracking when `FuncCallExpr` parse fails
2. backtracking in parenthesized expressions and tuples

Case 1 is fixed by re-writing the parse rule so that the function call target is relaxed to the superset of `Atom` expressions and then using obscure GrammarKit functionality (`left`) so that when we try to parse the target as an `Atom`, we can still use the parsed result even if it is not a function call (no whitespace separated arguments follow).

Case 2 backtracking can occur in:
- parenthesized expression `(4 + 3)`
- a tuple `(4, 3)`
- or a field access expression `(foo model).bar`

The problem is that the naive approach has to parse the expression following the opening `(` multiple times since we don't know what we are parsing until we reach a `,`, a `).` or a `)`. For simple examples like above, this backtracking is not that bad, but when it involves nested expressions, execution time goes non-linear. 

This PR fixes Case 1 and parts of Case 2. I still need to do more work to optimize how field access expressions are parsed.


klazuka:
Bummer... I forgot about pinning on the open paren (or, at a minimum, on the first comma in a tuple expression). I will add that before merging.

Herteby:
Woo! 😄🎉

Issue #344, opened 2019-03-07
PsiInvalidElementAccessException when commenting out a declaration


klazuka:
@clojj found that he can reproduce
`com.intellij.psi.PsiInvalidElementAccessException: Element: class org.elm.lang.core.psi.elements.ElmTypeAliasDeclaration #Elm  because: parent is null
`
as follows:
- open elm-spa-example (version without errors)
- open file `src/Page/Article/Editor.elm`
- comment out `type alias Model = ...` (using block-comment)

I confirmed that I can also reproduce it in the PowerLift project. So this appears to have been a major regression and should be fixed immediately.

klazuka:
At first I thought the problem was related to the `ResolveCache`, but I disabled it and the exception was still being thrown. 

One thing we could do is use `com.intellij.psi.PsiElement#isValid` in the various inspections to detect the condition and bail out before accessing the Psi element. It is done quite a bit throughout the IntelliJ codebase, although not so much in intellij-rust. Unclear whether it's the right thing to do, and I don't like how it's something that needs to be sprinkled around the code.

Another thing I don't understand is why this seems to only happen when commenting-out a declaration. I guess it's a timing thing? Maybe I need to handle the multithreading more carefully by explicitly acquiring a read action in more places?

klazuka:
Appears to have been fixed by #516

PR #516, opened 2019-09-23
Fix several cache invalidation bugs

ajalt:
- Language injections don't have an event system, so we need to invalidate their contents on any PSI change
- Invalidate globally when the body of an unannotated top-level function changes, since that can change the return type or parameters
- Invalidate globally when a VFS directory is deleted

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/psi/ElmPsiManager.kt
+++ b/src/main/kotlin/org/elm/lang/core/psi/ElmPsiManager.kt
@@ -58,7 +58,7 @@ class ElmPsiManager(val project: Project) : ProjectComponent {
                 modificationTracker.incModificationCount()
                 return
             }
-            
+
             if (file?.fileType != ElmFileType) return
             if (element is PsiComment || element is PsiWhiteSpace) return
 
--- a/src/main/kotlin/org/elm/lang/core/psi/ElmPsiManager.kt
+++ b/src/main/kotlin/org/elm/lang/core/psi/ElmPsiManager.kt
@@ -52,11 +52,14 @@ class ElmPsiManager(val project: Project) : ProjectComponent {
         }
 
         private fun onPsiChange(event: PsiTreeChangeEvent, element: PsiElement) {
-            // There are some cases when PsiFile stored in the event as a child
-            // e.g. file removal by external VFS change
-            val file = event.file ?: event.child as? PsiFile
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/elm/lang/core/psi/ElmPsiManager.kt
```diff
@@ -7,53 +7,73 @@
...
+        override fun propertyChanged(event: PsiTreeChangeEvent) {
+            if (event.propertyName != PROP_WRITABLE && event.element != null) {
+                onPsiChange(event, event.element)
+            }
+        }
+
+        private fun onPsiChange(event: PsiTreeChangeEvent, element: PsiElement) {
+            // If the file is null, then this is an event about VFS changes
+            val file = event.file
+            if (file == null && (element is ElmFile || element is PsiDirectory)) {
```
ajalt: If we have an easy way to check that a `PsiDirectory` is in an Elm source root, we could add that here as an optimization, but I don't think it's a big deal.

klazuka:
Nice. I've been testing it out and it looks like this change also fixed #344.

src/main/kotlin/org/elm/lang/core/psi/ElmPsiManager.kt
```diff
@@ -7,53 +7,73 @@
...
-            if (event is PsiTreeChangeEventImpl && event.isGenericChange) return
+            // We handle more concrete events and so should ignore the generic event.
+            if (event !is PsiTreeChangeEventImpl || !event.isGenericChange) onPsiChange(event, event.parent)
+        }
 
-            // There are some cases when PsiFile stored in the event as a child
-            // e.g. file removal by external VFS change
-            val file = event.file ?: event.child as? PsiFile
-            if (file?.fileType != ElmFileType) return
+        override fun propertyChanged(event: PsiTreeChangeEvent) {
```
klazuka: The javadoc on `PsiTreeChangeEvent` is pretty scary. But I guess we've been doing it for a while, so what can possibly go wrong 😄

src/main/kotlin/org/elm/lang/core/psi/ElmPsiManager.kt
klazuka: `org.elm.workspace.ElmWorkspaceService#findProjectForFile`. But I agree, it's probably not necessary.

PR #524, opened 2019-10-03
Ignore nested functions named 'main' when finding entry point

klazuka:
@dprophete reported on Slack that the Elm compiler panel was producing inconsistent/confusing results. The root cause was that he had a function with a `let/in` where it declared a nested function named `main` of type `Html`. Our code thought this nested function was `main`, sometimes.


dprophete:
Just noticed this. Thanks @klazuka

PR #551, opened 2019-11-19
Use Elm 0.19.1 for tests

ajalt:
In 0.19.1, it's now an error to compile a file without a module declaration. This change is not documented in the release notes or migration guide.


klazuka:
Good find! Thanks.

PR #566, opened 2019-12-18
Add filtered live template contexts

### LINKED ISSUE #543: Live templates included on comment lines
When typing certain characters (e.g. `t`) in a line which is a comment, the live templates for that character are shown. This isn't what happens in, say, Java or Kotlin code (no live templates are shown on comment lines).  For example in the gif below I type `-- t` then a live template shows up.  Additionally, if I now hit space (see note below) or tab, the live template will complete, and do something like add some code in the middle of my comment:

elm-live-template-on-comment

I wonder if all (or maybe just some) of the live templates should not be activated when writing on a comment line?

_Note_: hitting space completed the live template for me on one IntelliJ instance, but not another. Not sure which setting is at work here. Initially I thought it was "Insert selected suggestion by pressing space..." (shown below), but updating that setting didn't seem to have any impact so I'm not sure. Either way, I think that on a comment line in might be best for live templates not to be activated.

code-completion-settings

ajalt:
This allows us to only expand live contexts when they're syntactically valid.

Templates like `ty` and `mod` will now only expand at the top level in files.
Templates like `fn1` will now expand at top level or in a let declaration.
Templates like `case1` and `let1` will now only expand in an expression.

No templates will expand in a string or comment.

Fixes #543

### DIFF
--- a/src/test/kotlin/org/elm/ide/template/ElmLiveTemplatesTest.kt
+++ b/src/test/kotlin/org/elm/ide/template/ElmLiveTemplatesTest.kt
@@ -1,0 +1,212 @@
+package org.elm.ide.template
+
+import com.intellij.openapi.actionSystem.IdeActions
+import org.elm.lang.ElmTestBase
+import org.intellij.lang.annotations.Language
+
+class ElmLiveTemplatesTest : ElmTestBase() {
+    fun `test module`() = expandSnippet(
+            """
+mod{-caret-}
+""",
+            """
+module main exposing (..)
+
+
+""")
+
+    fun `test module in expression`() = noSnippet(
+            """
+main =
+  mod{-caret-}
+""")
+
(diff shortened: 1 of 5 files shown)


src/test/kotlin/org/elm/ide/template/ElmLiveTemplatesTest.kt
```diff
@@ -0,0 +1,212 @@
...
+       -> 
+          
+""")
+
+    private fun expandSnippet(@Language("Elm") before: String, @Language("Elm") after: String) =
+            checkByText(before, after) {
+                myFixture.performEditorAction(IdeActions.ACTION_EXPAND_LIVE_TEMPLATE_BY_TAB)
+            }
+
+    private fun noSnippet(@Language("Elm") code: String) = expandSnippet(code, code)
```
klazuka: clever!

PR #570, opened 2019-12-23
Make the elm-test code match our Kotlin style

klazuka:
There were some places where the Kotlin code was showing its roots from Java (or didn't match the way that most of our Kotlin code is written).


klazuka:
@frawa these changes should be pretty harmless (mostly just style), but if you have a few minutes to look it over, I'd appreciate it.

Issue #564, opened 2019-12-11
Can't set location of elm-test (installed with nvm)


yonigibbs:
I have elm, elm-test and elm-format all installed by nvm.  As mentioned in some other issues (#519 and #252) this means Auto Discover doesn't work, so I have to manually type in the paths.  This works for elm and elm-format, but no path I try seems to be accepted for elm-test:

![elm test path](https://user-images.githubusercontent.com/39593145/70630176-0d40b900-1c23-11ea-8149-6966550cc673.png)

In case it's relevant, when I run `which elm-test` I get: `/home/yoni/.nvm/versions/node/v10.16.0/bin/elm-test`.  I tried this path too, and get the same error as in the screenshot above.

yonigibbs:
Bit more information on this... looks like "Auto Discover" _does_ work, _if you install elm-test locally_.  In other words, if you have it installed globally with NVM, it doesn't seem to be found, but if you install it locally (e.g. `npm install elm-test --save-dev`) then it is found, in a path such as this:

    <your-project-path>/node_modules/.bin/elm-test

However the error above (the red text under the path to elm-test) still persists: `/usr/bin/env: 'node': No such file or directory`.

I've done some investigation and I think the cause of this is as follows...

The `elm-test` file which is auto-discovered has this content:

```
#!/usr/bin/env node
require("../lib/elm-test.js");
```

When a request is made to execute this file, the shebang line causes the system to look for `node` to execute this file, looking in the `$PATH` environment variable (as explained [here](https://stackoverflow.com/a/33510581/10326373)).  The problem is that the way the `elm-test` executable is run, the `$PATH` environment variable doesn't contain a path to `node`.  That's because the way the nvm install adds node to the path is by adding this content at the bottom of `.bashrc` (at least on my system that's how it did it):

```
export NVM_DIR="$HOME/.nvm"
[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
[ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completion
```

`node` is added to `$PATH` when the `.nvm` script is executed, which is controlled by `.bashrc`.  When command-line stuff is executed from Java, the `.bashrc` file isn't picked up unless the shell is invoked as an interactive shell, as explained [here](https://stackoverflow.com/a/47438096/10326373).  So that's why the error occurs.

[...]

PR #571, opened 2019-12-27
Added explanation of workaround for issue where plugin can't find elm…

yonigibbs:
Added explanation of workaround for issue where plugin can't find elm-format or elm-test, caused by node not being on the path, as discussed here: https://github.com/klazuka/intellij-elm/issues/564


yonigibbs:
Great, glad it was of use. Spotted one tiny typo in the changes so submitted #575 to fix it (just a missing trailing underscore).

PR #578, opened 2020-01-03
mention add missing case braches feature

pravdomil:


klazuka:
Thanks, but this would need to be like the other features where there is a separate Markdown document which describes the feature in more detail along with screenshots.

pravdomil:
I know, but this is better then not mention it at all

pravdomil:
its better to do not mention at all?

klazuka:
Yes. I'm trying to make sure that any feature which I talk about in the README has supporting documentation/screenshot(s).

Issue #563, opened 2019-12-11
Does folder containing tests have to be called "tests"?


yonigibbs:
If the folder containing tests isn't named `tests`, the code in IntelliJ goes red. Is there a setting I need to set somewhere?

This is shown in the screenshot below, where the folder is called `test` rather than `tests`:

![elm test folder](https://user-images.githubusercontent.com/39593145/70629431-e635b780-1c21-11ea-8d34-33fa6ad190f8.png)

If I rename the folder from `test` to `tests`, the red code goes black as normal.

The screenshot above is just a cut down example, where renaming the folder to "tests" would obviously be an easy answer.  However the reason I don't want a folder called "tests" is because in a real solution the structure is a bit more complicated (it's a Spring Boot app):

```
- src
  - main
    - kotlin
    - elm
  - test
    - kotlin
    - elm
      - <tests go here>
```

Note that I can run elm-test on a command-line and the tests run fine, regardless of the name of the folder.

yonigibbs:
I had a dig around the source code and I can see that this value ("tests") is hard-coded in a few places, e.g. `ElmProject`, `ElmModuleBuilder`, etc.  So looks like as the code currently stands the folder does indeed need to be called "tests".

I can also see that there is this comment in `ElmWorkspaceService`:

> Testing on Elm 0.19 requires that there be a "tests" directory next to the elm.json file. So we must register that as a source root if it exists.

Is that definitely true?  With the elm-test CLI you can specify the path to the test files, and that seems to work even if they're not in a folder called "tests".

If you agree it is possible for elm-test to work in a setup without a "tests" directory (which is how we've got it set up, and haven't noticed any problems so far), would it be possible to make the path to the folder containing the tests configurable?  Maybe a new value the user can enter in the Settings screen where they enter the location of elm-test?

klazuka:
@yonigibbs which version of elm-test are you using? I'm using 0.19.1-revision2, and I do not see an option to configure the location of the tests. When I rename my `tests` directory to, say, `foo`, and try to run `elm-test foo/*.elm`, I get the following error:
```
Error: /Users/keith/dev/elm-json-tree-view/tests does not exist. Please create a tests/ directory in your project root!
```

This makes it seem that elm-test only supports a directory named "tests".

yonigibbs:
I'm using the same version as you. I had to add an entry in `elm.json` under the `source-directories` key, e.g. if my tests are in `my-tests` then I have to do this:

```
{
[...]

PR #580, opened 2020-01-06
Configurable tests folder

yonigibbs:
Hi @klazuka. As discussed, here's the first part of the changes for #563. Functionally nothing has changed yet, but the code now allows the test folder location to be specified. To test this manually, see the TODO on line 158 of `ElmProject`: here you can manually specify a value such as "/src/test/frontend". In future this value will be set from a config file.

Some things to point out in this PR:
* In `ElmWorkspaceService` on line 318 the folder containing the tests is added as a source root. There's a comment saying this is done "... if it exists". There was no explicit check for the existence of the tests folder before the path was added: it looks like instead this is done in the `put` method a few lines higher, in the first couple of lines of that function (possibly where it calls `findFileByPathTestAware`). I decided to add the existence test explicitly before calling `put`, as I wasn't 100% sure about the inner workings of `findFileByPathTestAware`. Will remove if you think it’s not needed.
* I wasn't quite sure how to handle Elm 18. I guess the pragmatic decision might be to _not_ change any functionality here, and to leave it always using a folder called "tests". Especially as the PR we found in `node-test-runner` for making the folder configurable seems to have been made after Elm 19 was released, so I'm not sure that functionality would even be available on Elm 18. Let me know if there's a different approach you prefer.
* I haven't added many tests yet because, like mentioned above, functionally nothing has yet changed, and the system can't yet be configured to use a different tests folder. When we do the second bit of work to allow the user to specify the folder path, at that point I'll add a bunch of tests. We might then find more places where changes need to be made.
* I added a bit of logging in `ElmTestCLI` related to the change I made. This might be overkill. Will remove if you think this is redundant.

Cheers!

### DIFF
--- a/src/main/kotlin/org/elm/ide/test/core/LabelUtils.kt
+++ b/src/main/kotlin/org/elm/ide/test/core/LabelUtils.kt
@@ -50,6 +50,10 @@ object LabelUtils {
         return String.format("%s://%s", protocol, pathString(path))
     }
 
+    /**
+     * @param testsRelativeDirPath The path to the directory containing the tests, relative to the project's root (i.e.
+     * the folder containing `elm.json`). The supplied `path` is the location of the test file, relative to this directory.
+     */
     fun fromLocationUrlPath(path: String, testsRelativeDirPath: String): Pair<String, String> {
         val path1 = Paths.get(path)
         val moduleName = getModuleName(path1)
--- a/src/main/kotlin/org/elm/ide/test/core/LabelUtils.kt
+++ b/src/main/kotlin/org/elm/ide/test/core/LabelUtils.kt
@@ -50,10 +50,10 @@ object LabelUtils {
         return String.format("%s://%s", protocol, pathString(path))
     }
 
-    fun fromLocationUrlPath(path: String): Pair<String, String> {
+    fun fromLocationUrlPath(path: String, testsRelativeDirPath: String): Pair<String, String> {
         val path1 = Paths.get(path)
         val moduleName = getModuleName(path1)
-        val moduleFile = String.format("tests/%s.elm", moduleName.replace(".", "/"))
+        val moduleFile = "$testsRelativeDirPath/${moduleName.replace(".", "/")}.elm"
         val label = if (path1.nameCount > 1) decodeLabel(path1.subpath(1, path1.nameCount)) else ""
         return Pair(moduleFile, label)
     }
--- a/src/main/kotlin/org/elm/ide/test/run/ElmTestRunProfileState.kt
+++ b/src/main/kotlin/org/elm/ide/test/run/ElmTestRunProfileState.kt
@@ -89,9 +89,8 @@ class ElmTestRunProfileState internal constructor(
     }
 
     override fun createConsole(executor: Executor): ConsoleView? {
-        if (elmProject == null)
-            // Should never happen as we only start a test (causing this method to be called) if we found a project.
-            throw IllegalStateException("Elm project not found")
+        // Should always have an Elm project as we only start a test (causing this method to be called) if we found a project.
+        if (elmProject == null) throw IllegalStateException("Elm project not found")
 
         val runConfiguration = environment.runProfile as RunConfiguration
         val properties = ConsoleProperties(runConfiguration, executor, elmProject.testsRelativeDirPath)
--- a/src/main/kotlin/org/elm/ide/test/run/ElmTestRunProfileState.kt
+++ b/src/main/kotlin/org/elm/ide/test/run/ElmTestRunProfileState.kt
@@ -121,7 +121,8 @@ class ElmTestRunProfileState internal constructor(
 
         override fun createTestEventsConverter(testFrameworkName: String, consoleProperties: TestConsoleProperties): OutputToGeneralTestEventsConverter {
             return object : OutputToGeneralTestEventsConverter(testFrameworkName, consoleProperties) {
-                var processor = ElmTestJsonProcessor()
+                // TODO [tests-folder]: replace hard-coded string with value read from ElmProject.
+                var processor = ElmTestJsonProcessor("tests")
 
                 @Synchronized
                 override fun finishTesting() {
--- a/src/main/kotlin/org/elm/ide/test/run/ElmTestRunProfileState.kt
+++ b/src/main/kotlin/org/elm/ide/test/run/ElmTestRunProfileState.kt
@@ -121,8 +121,7 @@ class ElmTestRunProfileState internal constructor(
 
         override fun createTestEventsConverter(testFrameworkName: String, consoleProperties: TestConsoleProperties): OutputToGeneralTestEventsConverter {
             return object : OutputToGeneralTestEventsConverter(testFrameworkName, consoleProperties) {
-                // TODO [tests-folder]: replace hard-coded string with value read from ElmProject.
-                var processor = ElmTestJsonProcessor("tests")
+                var processor = ElmTestJsonProcessor(testsRelativeDirPath)
 
                 @Synchronized
                 override fun finishTesting() {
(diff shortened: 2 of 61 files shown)


klazuka:
Sorry for the delay. I have a cold, and I'm travelling for work. I'll review when I get a chance in the next few days.

src/main/kotlin/org/elm/ide/test/run/ElmTestRunProfileState.kt
```diff
@@ -83,14 +89,22 @@ class ElmTestRunProfileState internal constructor(
     }
 
     override fun createConsole(executor: Executor): ConsoleView? {
+        if (elmProject == null)
```
klazuka: just make this one line: `if (elmProject == null) error("Elm project not found")`

src/main/kotlin/org/elm/ide/test/core/LabelUtils.kt
```diff
@@ -50,10 +50,10 @@ object LabelUtils {
         return String.format("%s://%s", protocol, pathString(path))
     }
 
-    fun fromLocationUrlPath(path: String): Pair<String, String> {
+    fun fromLocationUrlPath(path: String, testsRelativeDirPath: String): Pair<String, String> {
```
klazuka: I'm not sure if you've explored this, but what if the "label" fully encoded the path to the Elm file so that we don't need to reconstruct it later?

For instance, in `org.elm.ide.test.core.ElmTestJsonProcessor#newTestSuiteStartedEvent` when we construct the location URL, we could embed the full, relative path to the Elm file instead of just the name of the module. Then you wouldn't need to make any changes in `ElmTestLocator` *and* it would make it easier in the future to handle multiple test roots.

Can you give that a try to see if it makes things simpler?

src/main/kotlin/org/elm/workspace/ElmWorkspaceService.kt
```diff
@@ -315,9 +315,9 @@ class ElmWorkspaceService(
                         // Elm 0.18 requires a separate manifest/project for the test code, so we
                         // don't need to register the 'tests' directory in this case.
                     } else {
-                        // Testing on Elm 0.19 requires that there be a "tests" directory next
-                        // to the elm.json file. So we must register that as a source root if it exists.
-                        put(project.manifestPath.resolveSibling("tests"), project)
+                        // Register the directory containing the tests as a source root if it exists.
+                        if (project.testsDirPath.exists())
```
klazuka: You don't need the existence check.

src/main/kotlin/org/elm/ide/test/core/LabelUtils.kt
yonigibbs: Funnily enough that was the first route I went down. The problem I found was that the path to a test is defined as a URL such as `elmTestTest://TheModule/some/other/stuff`.  Here, `TheModule` is the path to the test file (in this case "TheModule.elm"). If the file is in a subfolder the URL would be `elmTestTest://TheFolder.TheModule/some/other/stuff`. The important point here is that the name of the module can be extracted from the URL by looking in the first part of the path, as seen in `LabelUtils.fromLocationUrlPath`.  This module name is used, for example, when showing the test results in the treeview in IntelliJ.

Now, say the URL was expanded to include the full path to the file, such as this: `elmTestTest://src/test/frontend/TheFolder.TheModule/some/other/stuff`: the code wouldn't know exactly which part contained the module name.  So instead I think we'd have two options:
1. Store some parts(s) of the data in `metainfo` (one of the parameters received in `ElmTestLocation.getLocation`, which we can set, for example, in `ElmTestJsonProcess.newTestSuiteStatedEvent`: we just need to call a different constructor of `TestSuiteStartedEvent`, which receives the `metainfo`).  We'd have to do the same thing for `Test` as for `TestSuite`.
2. Come up with some sort of custom syntax that would allow us to extract the name of the module from the full file path. I've just noticed that a similar approach is already taken in `LabelUtils.kt` in the `ErrorLabelLocation` class, where it puts various data separated by `::`.

It was at this point of my investigations that I decided the other approach I eventually went with would be simpler, especially as I wanted to keep the PR smaller and more self-contained.

So, I totally agree with your suggested approach, and like I say it's the first thing I thought of too, but I wonder if it merits own separate PR?  What do you think?  And do you have a preference between the two options above, or any other ideas?

klazuka: I would prefer front-loading it, either by 
- encoding it in the label URL (using some kind of separator, as you suggested).
- or by changing the URL to be the full path to the file on disk and using the `metainfo` parameter for the Elm module name.

My preference would be for the `metainfo` option, but the first option might be less disruptive. The existing `LabelUtils` code is pretty gnarly, so it might be a pain either way.

In terms of how to go about it, I would keep this PR open and just create a new one to see how that comes out.

yonigibbs: All sounds good to me. I'll give it a bash and submit a PR when it's ready. Thanks!

yonigibbs: Spent a bit of time digging, and it seems to me that the least intrusive way of making this change would be to keep the URL as-is, and to store the path to the root of the `ElmProject` in the `metainfo`. That way we can combine the `metainfo` with the first part of the URL to get to the full path to the actual Elm file. Any other approach would require a larger-scale change I think, especially in the unit tests.

The JavaDocs about `metainfo` in `SMTestLocator` say this:

> The <code>metainfo</code> parameter simplifies the search for locations, but can not be used to identify the test. A good example for <code>metainfo</code> is the line number of the beginning of the test. It can speed up the search procedure, but it changes when editing.

So it's kinda debatable as to whether storing the location of the project in `metainfo` would be an advisable use it. I think you could probably argue it either way.

You happy for the PR to go down this route, for the sake of reducing the number of changes?  Or would you prefer a larger-scale refactor of this area of the code?

klazuka: Go with the simplest change. I'm not worried about "abusing" the `metainfo` parameter.

I didn't write the elm-test integration, and it has been my intention to rewrite it to match the way the rest of the plugin is written, but I just haven't had a chance to do it yet.

yonigibbs: #595 now implemented and merged into master, so I think this review comment can be resolved?  This PR has now been updated with all changes from master, including the changes in #595.

src/main/kotlin/org/elm/ide/test/run/ElmTestRunProfileState.kt
```diff
@@ -83,14 +89,21 @@ class ElmTestRunProfileState internal constructor(
     }
 
     override fun createConsole(executor: Executor): ConsoleView? {
+        // Should always have an Elm project as we only start a test (causing this method to be called) if we found a project.
+        if (elmProject == null) throw IllegalStateException("Elm project not found")
```
klazuka: No big deal, but I would use Kotlin's `error` function instead of throwing illegal state exception. And I would also get rid of the comment since I think it can be assumed.

src/main/kotlin/org/elm/workspace/ElmWorkspaceService.kt
```diff
@@ -315,9 +315,9 @@ class ElmWorkspaceService(
                         // Elm 0.18 requires a separate manifest/project for the test code, so we
                         // don't need to register the 'tests' directory in this case.
                     } else {
-                        // Testing on Elm 0.19 requires that there be a "tests" directory next
-                        // to the elm.json file. So we must register that as a source root if it exists.
-                        put(project.manifestPath.resolveSibling("tests"), project)
+                        // Register the directory containing the tests as a source root if it exists.
+                        // (Existence check done in `put` method above.)
```
klazuka: No big deal, but I would get rid of line 319 `// (Existence check done in 'put' method above.)`

yonigibbs:
Comments all sound great. Thanks! Will do the second part of the PR (which makes the folder configurable through a config file) over next couple of days when I get a chance.

klazuka:
Cool. I've been having a hard time keeping up with PRs this week. So no rush 😄

PR #585, opened 2020-01-10
Optimize reference resolve

ajalt:
This PR contains a number of micro and macro optimizations that improve the performance of reference resolve 50-500%, depending on the project. Larger projects see more benefit due to improved caching.

Please double check my cache invalidation, I believe I'm using the correct modification trackers, but it's worth another set of eyes. Also feel free to veto any of the changes if you feel they complicate the code too much.

Here are some of the larger changes:

## ModuleScope

Although these loops were already cached, they were cached as a list of elements. Most reference implementations that used this scope would copy the returned list to an array (in `getVariants`), then filter that array down the the element that matches their name (in `resolveInner`).

The elements are now cached as both an array and a map of name to element, so now we don't make any copies, and `resolveInner` can do an O(1) map lookup. As an additional benefit, we only look up the element's `referenceName` once, where we were previously looking it up for each element in the array being searched.

## ImportScope

These results are now cached in a form similar to the form used in `ModuleScope`.

More importantly, calculating the values has been improved. Previously, calculating exposed names involved resolving the reference of each name in the exposing list. To do so, we would walk up from the reference to the file, then call a `ModuleScope.getDeclared*` function and filter the list for a matching name. But since all the exposed elements are defined in the same file, we now only call `ModuleScope` once and filter all the declarations in one pass. 

## ExpressionScope

This scope is very hot, but isn't amenable to caching due to the fact that the tree walk is unique for each reference, and retrieving a cached value is slow enough that caching intermediate results (e.g. `ElmValueDeclaration.declaredNames`) is a pessimization.

The main improvement to this scope is making the tree walk lazy so that e.g. resolving a reference to a function parameter doesn't need to call `ModuleScope.getVisibleValues`.

## GlobalScope

Since these results are the same for all files in a project, the results are now cached on the `ElmProject`. Additionally, the loops benefit from the improvements to `ModuleScope`

## Microoptimizations

I'm not going to list every change here, but there are a number of places that would walk over the same part of the Psi tree repeatedly. For example, `ElmUpperCaseQID.fillName`, which is a hot function, would gather all it's children into a list three times (once each for `isQualified`, `qualifierPrefix`, and `refName`). Now that function just calls `text` if it isn't a stub.

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/resolve/scope/GlobalScope.kt
+++ b/src/main/kotlin/org/elm/lang/core/resolve/scope/GlobalScope.kt
@@ -1,16 +1,26 @@
 package org.elm.lang.core.resolve.scope
 
+import com.intellij.openapi.util.Key
+import com.intellij.psi.util.CachedValueProvider
+import com.intellij.psi.util.CachedValuesManager
+import com.intellij.psi.util.ParameterizedCachedValue
+import org.elm.lang.core.lookup.ClientLocation
 import org.elm.lang.core.psi.ElmFile
 import org.elm.lang.core.psi.ElmNamedElement
 import org.elm.lang.core.psi.elements.ElmModuleDeclaration
+import org.elm.lang.core.psi.globalModificationTracker
 import org.elm.lang.core.stubs.index.ElmModulesIndex
+import org.elm.workspace.elmWorkspace
 
+private val VISIBLE_VALUES_KEY: Key<ParameterizedCachedValue<List<ElmNamedElement>, ElmFile>> = Key.create("VISIBLE_VALUES_KEY")
+private val VISIBLE_TYPES_KEY: Key<ParameterizedCachedValue<List<ElmNamedElement>, ElmFile>> = Key.create("VISIBLE_TYPES_KEY")
+private val VISIBLE_CTORS_KEY: Key<ParameterizedCachedValue<List<ElmNamedElement>, ElmFile>> = Key.create("VISIBLE_CTORS_KEY")
 
 /**
  * The subset of implicitly exposed values, types and constructors provided by Elm's
  * standard library ("Core").
  */
-class GlobalScope private constructor(val clientFile: ElmFile) {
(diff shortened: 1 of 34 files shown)


src/main/kotlin/org/elm/lang/core/resolve/scope/GlobalScope.kt
```diff
@@ -68,57 +77,82 @@ class GlobalScope private constructor(val clientFile: ElmFile) {
...
+                ctors["Ok"]?.let { rest.add(it) }
+                ctors["Err"]?.let { rest.add(it) }
+            }
+            return rest
+        }
+    }
 
-        return rest
+    fun getVisibleValues(): List<ElmNamedElement> {
+        val elmProject = clientFile.elmProject ?: return produceVisibleValues(clientFile)
```
klazuka: If the `elmProject` property is null, just return early with an empty list. We cannot provide good results without it.

Ditto for `getVisibleTypes()` and `getVisibleConstructors()`

src/main/kotlin/org/elm/lang/core/resolve/scope/ImportScope.kt
```diff
@@ -1,12 +1,87 @@
...
+    }
+
+    fun getExposedConstructors(): Sequence<ElmNamedElement> {
+        return scopes().flatMap { it.getExposedConstructors().elements.asSequence() }
+    }
+
+    /** Lazily yield all individual scopes that can contain exposed names with this prefix */
+    private fun scopes(): Sequence<ImportScope> = sequence {
+        if (importsOnly) {
+            yieldAll(implicitScopes())
```
klazuka: Shouldn't the explicit scopes be yielded first? I see that this is how the existing code was written, but it's the opposite of what we do in `VisibleNames.getAll()` where first we have top-level decls, then imports, then implicit/global. I haven't been able to produce an example where your way is wrong, but for consistency and sanity, I think it makes sense if explicit comes first.

If you do make this change, we should also change `allScopes` on line 74 to match.

PR #591, opened 2020-01-13
Add options to disable individual line marker types

### LINKED ISSUE #590: Can't configure elm gutter icons
there's no options for controling elm gutter icons eg. exposed?

image

ajalt:
This consolidates all line marker providers into a single `LineMarkerProviderDescriptor` implementation that calls out to the individual providers and handles registering and checking options, as well as the boilerplate for `collectSlowLineMarkers`. Individual providers can just implement `getLineMarkerInfo`. Providers are created for each call to `collectSlowLineMarkers`, so they can have internal state if they need to.

Fixes #590

Capture

### DIFF
--- a/src/main/kotlin/org/elm/ide/lineMarkers/ElmLineMarkerProvider.kt
+++ b/src/main/kotlin/org/elm/ide/lineMarkers/ElmLineMarkerProvider.kt
@@ -1,7 +1,6 @@
 package org.elm.ide.lineMarkers
 
 import com.intellij.codeInsight.daemon.LineMarkerInfo
-import com.intellij.codeInsight.daemon.LineMarkerProvider
 import com.intellij.codeInsight.daemon.LineMarkerProviderDescriptor
 import com.intellij.openapi.progress.ProgressManager
 import com.intellij.openapi.project.DumbService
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/elm/ide/lineMarkers/ElmLineMarkerProvider.kt
```diff
@@ -0,0 +1,41 @@
...
+    override fun getOptions(): Array<Option> = OPTIONS
+
+    override fun collectSlowLineMarkers(elements: MutableList<PsiElement>, result: MutableCollection<LineMarkerInfo<PsiElement>>) {
+        val first = elements.firstOrNull() ?: return
+        if (DumbService.getInstance(first.project).isDumb || first.parentOfType<ElmFile>()?.elmProject == null) return
+
+        // Create the providers each time this function is called, since they may have internal state
+        val providers = mutableListOf<LineMarkerProvider>()
+        if (ElmExposureLineMarkerProvider.OPTION.isEnabled) providers.add(ElmExposureLineMarkerProvider())
+        if (ElmRecursiveCallLineMarkerProvider.OPTION.isEnabled) providers.add(ElmRecursiveCallLineMarkerProvider())
```
klazuka: It seems like it would be easy to add a new provider here but forget to also update `OPTIONS` above. Maybe augment `OPTIONS` so that each element knows how to construct the corresponding provider instance? Then you could just loop over `OPTIONS` to create the list of providers.

If that makes the code too complicated, then just add a comment on line 31 that calls out the need to update `OPTIONS` manually.

ajalt: 👍 That's a great idea

PR #594, opened 2020-01-14
Fixed compilation problem caused by last commit.

yonigibbs:


ajalt:
Thanks! I should have rebased my branch after this inspection was added.

PR #595, opened 2020-01-14
Folder containing tests no longer hard-coded in LabelUtils...

yonigibbs:
... Now passed in as parameter from ElmTestJsonProcessor, and stored in metainfo.

This is the discussed pre-requisite for #580, which means that the `metainfo` for a location URL now contains the path to the folder containing the tests. The Elm module specified in the first portion of the URL is relative to the folder in `metainfo`. Currently it's hard-coded to "tests": that will change if/when we proceed with the follow-up to #580 which will allow this value to specified in some config file.

The main changes in this PR are as follows:
* When `ElmTestJsonProcessor` constructs a `TestStartedEvent` and `TestSuiteStartedEvent` it now passes in the path to the tests folder in the `metainfo` argument. In order to do this, a couple of methods that were previously static are now instance methods, so that they can access the `testsRelativeDirPath` property. The content of these methods hasn't changed (other than passing in `metainfo` where relevant).
* When `ElmTestLocator` calls `LabelUtils.fromLocationUrlPath` it now passes is the `metainfo`, so that `LabelUtils` can figure out the full path to the file: previously it assumed the file was in a folder called "tests".


yonigibbs:
Ah, I'm using `Paths.get` and it's getting backslashes on Windows instead of forward-slashes as the test code assumes. Leave this PR with me and I'll look at it and resolve soon as I get a chance.

yonigibbs:
Fixed the issue the was causing the PR to fail the CI build. Now ready for review.

src/main/kotlin/org/elm/ide/test/run/ElmTestLocator.kt
```diff
@@ -32,7 +34,9 @@ object ElmTestLocator : FileUrlProvider() {
             }
 
             DESCRIBE_PROTOCOL, TEST_PROTOCOL -> {
-                val (filePath, labels) = LabelUtils.fromLocationUrlPath(path)
+                // TODO [tests-folder]: replace hard-coded string with constant defined in `ElmProject` when that change
+                //  is merged in.
+                val (filePath, labels) = LabelUtils.fromLocationUrlPath(path, metainfo ?: "tests")
```
klazuka: I would get rid of the fallback when `metainfo` is null. The fallback will not always be correct, so I'd rather that it just blows up. e.g. `error("missing path to tests dir")`

yonigibbs: Agreed. Done.

PR #597, opened 2020-01-15
Support configurable indent size

### LINKED ISSUE #567: Tab size
Hi Keith. Thanks for this great plugin! 

Is there an option to control tab size? As far as I know the Elm community is not settled on 2 vs 4 spaces. So we need a toggler. I see that tab size (number of spaces) is inherited from "Default Language". It's good but it's not enough and some behaviors ignore that default. For example:

```elm
myfunction : Int -> Int
myfunction = -- press Enter
    _ -- gets here no matter of actual tab size
```

ajalt:
This PR adds support for a very basic Elm code style: we allow only a custom indent size. The enter handlers and intentions that generate code were updated to check the code style.

Fixes #567

### DIFF
--- a/src/main/kotlin/org/elm/ide/inspections/MissingCaseBranchAdder.kt
+++ b/src/main/kotlin/org/elm/ide/inspections/MissingCaseBranchAdder.kt
@@ -1,5 +1,6 @@
 package org.elm.ide.inspections
 
+import com.intellij.application.options.CodeStyle
 import com.intellij.psi.PsiDocumentManager
 import org.elm.lang.core.lookup.ElmLookup
 import org.elm.lang.core.psi.ElmPsiFactory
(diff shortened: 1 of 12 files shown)


src/main/kotlin/org/elm/ide/typing/ElmOnEnterIndentHandler.kt
```diff
@@ -35,7 +36,7 @@ import org.elm.lang.core.psi.elements.ElmTypeDeclaration
  * "smart" indentation to make it easy to write Elm code, especially
  * since Elm's parser depends on indentation as an implicit delimiter.
  */
-class ElmOnEnterSmartIndentHandler : EnterHandlerDelegateAdapter() {
+class ElmOnEnterIndentHandler : EnterHandlerDelegateAdapter() {
```
ajalt: A "smart enter handler" has an existing meaning, so I renamed this to avoid confusion. I can undo the change if you'd like.

klazuka: 👍

src/main/kotlin/org/elm/ide/inspections/MissingCaseBranchAdder.kt
```diff
@@ -52,10 +53,11 @@ class MissingCaseBranchAdder(val element: ElmCaseOfExpr) {
     }
 
     private fun addPatterns(patterns: List<String>) {
+        val indent = " ".repeat(CodeStyle.getIndentOptions(element.containingFile).INDENT_SIZE)
```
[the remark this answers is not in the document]
ajalt: Yeah, that's definitely better.

PR #601, opened 2020-01-19
Added concept of custom manifest (elm.intellij.json)

yonigibbs:
This file is used to define the path to the directory containing the tests.

Some things to point out for your consideration when reviewing:

* I've gone with `elm.intellij.json` as the name of the file, so it's easy to find as it should always be right next to `elm.json`.
* In the code `elm.json` is referred to as the "manifest" so in terms of naming variables etc I came up with the name **"custom manifest"** for this new file.
* As mentioned in a previous PR, this is for Elm 19+ only, so for Elm 18 nothing should change.
* Am treating invalid JSON in the custom manifest as an error in the same way as invalid JSON in the normal manifest.  (One slight difference here is that I didn't put the `@JsonIgnoreProperties(ignoreUnknown = true)` annotation on the DTO for the custom manifest, as it should only ever have JSON attributes that we specifically define.)
* I slightly refactored some code in a couple of existing test classes: nothing functional should have changed in these.  The only reason I did this was to avoid a fair amount of duplication I'd have had to have done in my new tests otherwise.
* I added a couple of tests to `ElmWorkspaceResolveTest`: this might be testing too many “units” in one test, so you might want me to change this.  Possibly the tests in `ElmWorkspaceServiceTest` are enough.
* From what I could see there were no existing unit tests of `ElmTestCLI`. Is this more of an issue now that the folder path is configurable?

I know you're a bit snowed under just now with PRs so please take your time, absolutely no rush.


yonigibbs:
Apologies for the couple of failed CI builds (and hence three commits instead of one).  Windows paths :-(  It might be worth me spending some time setting up IntelliJ on Windows so I can run the unit tests locally on Windows as well as Linux before submitting a PR in future.

klazuka:
No problem! Thanks for contributing. I appreciate the PR summary, and I think that all of the considerations you made were sensible. 

The only thing I would change is the name "custom manifest". The way I see it, this is an extra layer on top of the normal manifest as opposed to an alternative to the normal manifest file. So I would try to go for a name that suggests the "additive" quality. Some considerations:

- addendum manifest
- appendix manifest
- overlay manifest

For this case where we are correcting for an oversight in the normal `elm.json` file, I think something like "addendum" would make a lot of sense since it corrects an error and is supplementary. And ultimately the goal should be to get rid of this extra manifest file once the `elm.json` file format is improved.

It's hard for me to imagine the names without seeing it in the codebase, so I'm open to other suggestions, too.

klazuka:
As for workspace tests, yeah, I know, it's a mess. I just did the minimum necessary to get the proper test coverage. My thinking was that once I get rid of Elm 0.18 support, I could refactor it to be nicer. But for now it's a mess. So if you can find a way to refactor to avoid duplication, that would be great. Otherwise, don't worry about it.

docs/features/elm-test.md
```diff
@@ -23,3 +23,27 @@ Once you have configured the path to `elm-test` (see above), right-click anywher
...
+    ```
+    {
+      "test-directory": "..."
+    }
+    ```
+   The value for `test-directory` should be the path to the folder containing the unit tests, relative to `elm.json`,
+   e.g. `src/test/elm`.
+   This folder can contain subfolders - all tests anywhere under this folder will be executed.
+
+1. Update `elm.json` and add the folder to the `source-directories` field.
```
klazuka: I would make this a little more concrete:

"...and add the value you use used for `"test-directory"` above to the `source-directories` array."

yonigibbs:
Cool, all sounds good.  I'll update the code to go with `addendum` instead of `custom` and will push the changes to this PR when I'm done.  The minor refactors in the test classes are pretty trivial: just wanted to let you know about them before you dive into the review.  Thanks!

src/main/kotlin/org/elm/workspace/ElmProject.kt
```diff
@@ -333,6 +356,19 @@ private class ElmApplicationProjectDTO(
...
+/**
+ * DTO used to wrap the data in `elm.intellij.json`.
+ *
+ * @see [ElmToolchain.ELM_INTELLIJ_JSON]
+ */
+private class ElmCustomManifestDTO(
+        /**
+         * The path to the directory containing the unit tests, relative to the root of the Elm project.
+         */
+        @JsonProperty("test-directory") val testDirectory: String?
```
klazuka: Why make this String nullable?

src/test/kotlin/org/elm/workspace/ElmWorkspaceServiceTest.kt
```diff
@@ -545,6 +476,88 @@ class ElmWorkspaceServiceTest : ElmWorkspaceTestBase() {
         val actualXml = workspace.state.toXmlString()
         checkEquals(xml, actualXml)
     }
+
+
+    // START OF TESTS RELATED TO CUSTOM MANIFEST (elm.intellij.json)
+
+    fun `test uses default tests location when no custom manifest exists (application project)`() =
```
klazuka: Nice. I like how you did the tests here.

src/test/kotlin/org/elm/workspace/ElmWorkspaceServiceTest.kt
```diff
@@ -13,21 +14,7 @@ class ElmWorkspaceServiceTest : ElmWorkspaceTestBase() {
...
-                        "direct": {},
-                        "indirect": {}
-                      },
-                      "test-dependencies": {
-                        "direct": {},
-                        "indirect": {}
-                      }
-                    }
-                    """)
+                project("elm.json", BASIC_APPLICATION_MANIFEST)
```
klazuka: 👍

src/main/kotlin/org/elm/workspace/ElmProject.kt
yonigibbs: Yeah, good point. I was thinking that in future we might add more to this file, and the user might want to set the other fields that we might add, but not to set this one, so the DTO should allow this to be null.  But then... YAGNI 😄 .  Will make it non-nullable!

klazuka:
After thinking about it more, I think we should call it a “sidecar” manifest. It’s more evocative of the purpose.

yonigibbs:
All requested changes now made. Let me know if you want me to mark any of the conversations above as resolved.

BTW the first CI build failed with that same intermittent error: `elm-stuff/0.19.1/d.dat: openBinaryFile: resource busy (file is locked)`.  So I just added a second commit with nothing in it to kick off the CI build again. Second time lucky. I take it there's no other way force the CI build to retry? I think when I get some time I'm going to see if I can find a way to replicate this more consistently so maybe we can get to the bottom of it.

PR #602, opened 2020-01-22
Optimize type inference

ajalt:
This wraps up my optimization works started in #585, with a focus on the type inference code. Inference in general spends > 60% of its time resolving references, so there isn't a huge opportunity for speedups. Most of the remaining 40% is spent in type replacement, so that's where I focused my effort.

In total, this PR speeds up total inference time by around 10%, but more importantly decreases the memory usage of the inference cache by about half.

The speedups mostly comes from reducing redundant tree walking and redundant type replacement copying.

The memory reduction is due partially to reducing copies in type replacement, allowing more shared references, but mostly by no longer allocating any extra capacity in the cached data structures. 

While working on this, I found a couple of minor bugs in inference that I've fixed and added tests for.

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/types/RecordFieldReferenceTable.kt
+++ b/src/main/kotlin/org/elm/lang/core/types/RecordFieldReferenceTable.kt
@@ -1,13 +1,25 @@
 package org.elm.lang.core.types
 
+import com.intellij.util.containers.SmartHashSet
 import org.elm.lang.core.psi.ElmNamedElement
+import org.elm.lang.core.psi.elements.ElmFieldType
 
 /**
  * A table that tracks references for [TyRecord] fields. Can be [frozen] to prevent updates.
  */
-class RecordFieldReferenceTable(
-        private var refsByField: MutableMap<String, MutableSet<ElmNamedElement>> = LinkedHashMap(4, 1f)
+class RecordFieldReferenceTable private constructor(
+        private var refsByField: MutableMap<String, MutableSet<ElmNamedElement>>
 ) {
+    constructor() : this(HashMap(4, 1f))
+
+    companion object {
+        fun fromElements(fieldElements: Collection<ElmFieldType>): RecordFieldReferenceTable {
+            return RecordFieldReferenceTable(fieldElements.associateTo(HashMap(fieldElements.size, 1f)) {
+                it.name to newSet(1).apply { add(it) }
+            })
+        }
+    }
+
     var frozen = false
         private set
 
@@ -20,7 +32,9 @@ class RecordFieldReferenceTable(
     fun addAll(other: RecordFieldReferenceTable) {
         if (frozen || other.refsByField === this.refsByField) return
         for ((field, refs) in other.refsByField) {
-            refsByField.getOrPut(field) { HashSet(refs.size, 1f) } += refs
+            // Don't use putAll here, since that function will resize the table to hold (size + 1) elements
+            val set = refsByField.getOrPut(field) { newSet(refs.size) }
+            refs.forEach { set.add(it) }
         }
     }
 
@@ -29,13 +43,15 @@ class RecordFieldReferenceTable(
 
     /** Create a new table with references from this table and [other] */
     operator fun plus(other: RecordFieldReferenceTable): RecordFieldReferenceTable {
-        val newRefs = LinkedHashMap<String, MutableSet<ElmNamedElement>>(refsByField.size + other.refsByField.size, 1f)
+        val initialCapacity = refsByField.size + other.refsByField.keys.count { it !in refsByField }
+        val newRefs = HashMap<String, MutableSet<ElmNamedElement>>(initialCapacity, 1f)
         refsByField.mapValuesTo(newRefs) { (field, set) ->
             val otherSet = other.refsByField[field].orEmpty()
-            HashSet<ElmNamedElement>(set.size + (otherSet.size), 1f).apply { addAll(otherSet) }
+            val initialSetCapacity = set.size + otherSet.count { it !in set }
+            newSet(initialSetCapacity).apply { otherSet.forEach { add(it) } }
         }
         other.refsByField.forEach { (k, s) ->
-            newRefs.getOrPut(k) { HashSet<ElmNamedElement>(s.size, 1f).apply { addAll(s) } }
+            newRefs.getOrPut(k) { newSet(s.size).apply { s.forEach { add(it) } } }
         }
         return RecordFieldReferenceTable(newRefs)
     }
@@ -48,4 +64,9 @@ class RecordFieldReferenceTable(
     override fun toString(): String {
         return "<RecordFieldReferenceTable frozen=$frozen, refs=${refsByField.keys}>"
     }
+}
+
+private fun newSet(initialCapacity: Int): MutableSet<ElmNamedElement> {
+    // > 99% of these sets have 1 element, so we use a set optimized for that use case
+    return if (initialCapacity <= 1) SmartHashSet(-1, 1f) else SmartHashSet(initialCapacity, 1f)
 }
--- a/src/main/kotlin/org/elm/lang/core/types/RecordFieldReferenceTable.kt
+++ b/src/main/kotlin/org/elm/lang/core/types/RecordFieldReferenceTable.kt
@@ -6,7 +6,7 @@ import org.elm.lang.core.psi.ElmNamedElement
  * A table that tracks references for [TyRecord] fields. Can be [frozen] to prevent updates.
  */
 data class RecordFieldReferenceTable(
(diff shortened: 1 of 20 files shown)


src/main/kotlin/org/elm/lang/core/types/RecordFieldReferenceTable.kt
```diff
@@ -27,17 +43,30 @@ data class RecordFieldReferenceTable(
...
+        return RecordFieldReferenceTable(newRefs)
     }
 
-    fun copy(frozen: Boolean = this.frozen): RecordFieldReferenceTable {
-        return RecordFieldReferenceTable(
-                refsByField.mapValuesTo(mutableMapOf()) { (_, v) -> v.toMutableSet() },
-                frozen
-        )
+    /** Prevent further modifications to this table. Tables cannot be unfrozen. */
+    fun freeze() {
```
ajalt: Reference tables are now frozen in place, avoiding copies just to freeze them. This mostly only helps in the type expression inference, since the freeze was done in the same pass as replacement for value expressions. Even in that case, this can reduce copies if no other replacements need to be made to part of a ty.

src/main/kotlin/org/elm/lang/core/types/Ty.kt
```diff
@@ -224,34 +230,42 @@ object TyInProgressBinding : Ty() {
...
-/** Create a lazy sequence of all [TyVar]s referenced within this ty. */
-fun Ty.allVars(includeAlias: Boolean = false): Sequence<TyVar> = traverse(includeAlias).filterIsInstance<TyVar>()
+/** Return a list of all [TyVar]s in this ty, including itself */
+fun Ty.allVars(): List<TyVar> = mutableListOf<TyVar>().also { allVars(it) }
 
-fun Ty.traverse(includeAlias: Boolean = false): Sequence<Ty> = sequence {
-    yield(this@traverse)
-    when (this@traverse) {
-        is TyTuple -> types.forEach { yieldAll(it.traverse(includeAlias)) }
+private fun Ty.allVars(result: MutableList<TyVar>) {
```
ajalt: This code was very hot, so I replaced the coroutine-based `traverse` with hand-coded recursive descent for each use-case. Surprisingly, the coroutines were only ~25% slower, so maybe the extra code isn't worth it.

src/main/kotlin/org/elm/lang/core/types/TypeReplacement.kt
```diff
@@ -77,38 +71,63 @@ class TypeReplacement(
...
+            ty.alias?.parameters?.forEach { freeze(it) }
         }
+    }
 
-        private fun Ty.isUnfrozenRecord() = when (this) {
-            is TyRecord -> !fieldReferences.frozen
-            is MutableTyRecord -> !fieldReferences.frozen
-            else -> false
-        }
+    private fun wouldChange(ty: Ty): Boolean {
```
ajalt: Because this is now a member method, I experimented with caching the result in an `IdentityHashMap` to avoid repeated traversal when walking down the ty for replacement. This didn't have any performance impact. The cache was cheap to look up, but was almost always a miss, since tys are generally shallow.

src/main/kotlin/org/elm/lang/core/types/TypeReplacement.kt
```diff
@@ -77,38 +71,63 @@ class TypeReplacement(
...
+        return ty.f()
     }
 
     /** A map of var to (has been accessed, ty) */
     private val replacements = replacements.mapValuesTo(mutableMapOf()) { (_, v) -> false to v }
 
     fun replace(ty: Ty): Ty {
-        val replaced = when (ty) {
+        // If we wouldn't change anything, return the original ty to avoid duplicating the object
+        if (!wouldChange(ty)) return ty
```
ajalt: Besides the obvious improvement of not performing a copy then throwing it away if it was equal to the original, this also lets of get rid of using ty equality at all. I never liked using equality, since it was the source of a couple of hard to track down bugs.

src/main/kotlin/org/elm/lang/core/types/TypeReplacement.kt
```diff
@@ -159,21 +176,33 @@ class TypeReplacement(
...
-        val baseFields = (replacedBase as? TyRecord)?.fields.orEmpty()
-        val baseFieldRefs = (replacedBase as? TyRecord)?.fieldReferences
+        val baseFields = (oldBase as? TyRecord)?.fields.orEmpty()
+        val baseFieldRefs = (oldBase as? TyRecord)?.fieldReferences
 
-        var newFieldReferences = when {
+        // Make the new map as small as we can for these fields. HashMap always uses a power of two
+        // element array, so we can't avoid all wasted memory.
+        val initialCapacity = baseFields.size + fields.keys.count { it !in baseFields }
+        val newFields = LinkedHashMap<String, Ty>(initialCapacity, 1f)
```
ajalt: Even though `LinkedHashMap` uses a little extra memory, I think it's worth it to keep fields in the order that they're defined.

klazuka:
Cool. Today I'm going to publish a release based on current master. Then I'm going to be on vacation for a couple of days. When I get back I'll review this PR.

src/main/kotlin/org/elm/lang/core/types/TypeReplacement.kt
```diff
@@ -159,21 +176,33 @@ class TypeReplacement(
...
+
+        // Don't use putAll here, since that function will resize the table to hold (size + 1) elements
```
klazuka: You had a similar comment for `RecordFieldReferenceTable`. Maybe it's not possible due to different types, but would it be possible to define an extension method that makes the code self-documenting and reusable in both places?

klazuka:
Looks good. Just had a few minor questions. It looks like the optimizations didn't pollute the code too much (and in some places it seems simpler). Sorry for the long delay in review.

src/main/kotlin/org/elm/lang/core/types/RecordFieldReferenceTable.kt
```diff
@@ -27,17 +43,30 @@ data class RecordFieldReferenceTable(
...
     }
 
     override fun toString(): String {
         return "<RecordFieldReferenceTable frozen=$frozen, refs=${refsByField.keys}>"
     }
 }
+
+private fun newSet(initialCapacity: Int): MutableSet<ElmNamedElement> {
+    // > 99% of these sets have 1 element, so we use a set optimized for that use case
+    return if (initialCapacity <= 1) SmartHashSet(-1, 1f) else SmartHashSet(initialCapacity, 1f)
```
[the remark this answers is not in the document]
ajalt: I use the explicit constructor so that we have a default capacity with a load factor of 1. I agree it's non-obvious. I'll add a comment.

src/main/kotlin/org/elm/lang/core/types/Ty.kt
```diff
@@ -224,34 +230,42 @@ object TyInProgressBinding : Ty() {
...
         }
-        is TyUnion -> {
-            parameters.forEach { yieldAll(it.traverse(includeAlias)) }
-        }
-        is TyVar, is TyUnit, is TyUnknown, TyInProgressBinding -> {
+        is TyUnit, is TyUnknown, TyInProgressBinding -> {
         }
     }
-    if (includeAlias) {
-        alias?.parameters?.forEach { yieldAll(it.traverse(includeAlias)) }
```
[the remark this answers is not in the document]
ajalt: Yes. The only place this function is called now doesn't use them.

src/main/kotlin/org/elm/lang/core/types/TypeReplacement.kt
```diff
@@ -77,38 +71,63 @@ class TypeReplacement(
...
+    }
 
-        private fun Ty.isUnfrozenRecord() = when (this) {
-            is TyRecord -> !fieldReferences.frozen
-            is MutableTyRecord -> !fieldReferences.frozen
-            else -> false
-        }
+    private fun wouldChange(ty: Ty): Boolean {
+        val changeAllVars = freshen || varsToRemainRigid != null
+        fun Ty.f(): Boolean = when (this) {
```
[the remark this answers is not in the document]
ajalt: 🙏

src/main/kotlin/org/elm/lang/core/types/RecordFieldReferenceTable.kt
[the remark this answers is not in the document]
ajalt: Actually, if we use a positive initial capacity, we could use `addAll` on these sets here without causing a rehash in the case of single element sets.

src/main/kotlin/org/elm/lang/core/types/TypeReplacement.kt
ajalt: They're different types, so we'd need to functions. But it turns out that we can use `addAll` on a `SmartHashSet` after all, so I'll remove the comment there.

PR #620, opened 2020-03-02
Improve unresolved reference fixes

### LINKED ISSUE #399: Auto-import does the wrong thing when custom type and variant have the same name
reported by Simon:

if you try to autoimport a type constructor, it only gets imported as `import Toop exposing (T4)`, rather than `import Toop exposing (T4(..))`

which if you then try to use it to create something, causes the somewhat confusing compiler error `I cannot find a T4 constructor:`

it's only when the type and the variant have the same name though
ie. if you have a module which exposes:
```
type Test1
    = Test1


type Test2
    = Test3
```

and in another module you try

```
test =
    Test1


test2 =
    Test3
```

and attempt to autoimport them, `Test1` fails but `Test3` works

### LINKED ISSUE #599: Import fix tooltip
In TypeScript there is sligtly different import dialog, that can import missing modules more quickly (you don't need to press enter twice and there is option to import them all at one), is that possible in Elm too?

**elm**
image

**typescript**
image

### LINKED ISSUE #606: Import fix options
**First case**
I have a `type Model` defined in 3 modules: `Utils.One, Utils.Two, Three`.

And now I'm in Main.elm and I want to import `One.Model`. It offers module import `Three as One` as a first option, but it's obvious that I want to import `Utils.One as One`.

image

**Second case**
Why import `Json.Encode as Decode`?

image

**Question**
Can I make it stop doing that? So if I have `Foo.Bar.Model` it can only offer modules those names ends with `Foo.Bar`? And not all modules that have `Model` defined?

### LINKED ISSUE #607: Import fix offers type in place where can be only value
**Utils.UserId**
image

**Main.elm**
image

It doesn't make sense to offer `import Utils.UserId exposing (UserId)`.

ajalt:
- Only import type constructors in value expressions and only variants in type expressions. 
- Show a hint popup to instantly add an import or qualifier
- Improve candidate sorting when multiple modules could be imported 

Fixes #399
Fixes #599
Fixes #607
Fixes #606


src/main/kotlin/org/elm/ide/inspections/ElmUnresolvedReferenceInspection.kt
```diff
@@ -50,10 +50,31 @@ class ElmUnresolvedReferenceInspection : ElmLocalInspection() {
...
-            val description = "Unresolved reference '${ref.canonicalText}'"
             val fixes = mutableListOf<LocalQuickFix>()
-            if (AddImportFix.isAvailable(element)) fixes += AddImportFix()
-            if (AddQualifierFix.isAvailable(element)) fixes += AddQualifierFix()
+            val qualifierContext = AddQualifierFix.findApplicableContext(element)
+            val importContext = AddImportFix.findApplicableContext(element)
+            // Share the tracker between both fixes so that invoking one will hide the hint from the other
+            val invocationTracker = QuickFixInvocationTracker()
+
+            // Only show the hint for the qualifier fix if both are available.
```
ajalt: You can't show multiple hints at the same time. If we wanted to provide access to both actions in a hint, we would have to combine the quick fixes and show both actions in the dropdown. But this seems like the action that will be most common.

klazuka:
I tested it out in [elm-spa-example](https://github.com/rtfeldman/elm-spa-example) `src/Page/Article.elm` and ran into a few problems.

On line 20 (`import Json.Decode as Decode`) I put the cursor at the end of the line and pressed backspace repeatedly. Once the `as` token is reached, some of the type references in the `Model` definition on line 38 go red (which I think is due to broken parse? maybe not a regression). If you then keep pressing backspace the errors will go away but there will still be a blue popup on line 39 saying `Session exposing (Session)?`. I would expect that to go away as it is already imported (seems like a UI state bug).

Also on line 20, if I delete the `as Decode` part, and then scroll down to line 522, it (correctly) gives me the popup to fix the import. Now trigger the fix and noticed that it incorrectly changes line 20 to read `import Json.Decode as Decode exposing ()` (it should not have added any "exposing" in this case).

ajalt:
Thank you for testing. The bugs you found exist on master, but it's good to get them fixed. If you edit the import list, the popup can remain around until the the next time the import inspection runs, which on my machine can take half a second or so. I think that's just how the system works; I don't know if there's a way to improve it.

klazuka:
🤦‍♂ Bummer! Nevermind then 😊 
I'll finish reviewing tomorrow.

klazuka:
It sounds like tracking the lifecycle of the hint popup is a pain. I noticed a case where the hint is still shown after a previous fix resolved multiple refs.

e.g.
```
foo = Json.Decode.map identity Json.Decode.string
```
When I activate the hint for `Json.Decode.map`, it inserts the import, and the unresolved ref error goes away for both the call to `map` and to `string` *but* a new hint is shown for `string` even though it's no longer necessary.

The hint goes away when I move the mouse, so it's not so bad, but it could be a little confusing. Any idea if this would be easy to fix? Maybe we need to notify IntelliJ after a fix is committed so that it can refresh hints?

ajalt:
Ok, instead of trying to track modifications ourselves, I changed the hint class to use the global modification tracker, which works across hints. Simpler and it should work in all cases.

pravdomil:
I think that this pr introduces https://github.com/klazuka/intellij-elm/issues/643

Issue #631, opened 2020-03-17
Conflicting declaration false positive


pravdomil:
I have been testing #622, and it will report `rect` as duplicated in this code:

```elm
import Svg exposing (rect)

foo rect =
    let
        bar =
            1
    in
    ()
```

<img width="353" alt="image" src="https://user-images.githubusercontent.com/2387356/76905815-01f4cd80-68a3-11ea-9e42-911e62aad1df.png">

But compiler is ok with it.

pravdomil:
this maybe related

```
com.intellij.diagnostic.PluginException: Reported element PsiElement(LOWER_CASE_IDENTIFIER) is not from the file 'xxx.elm' the inspection 'ElmDuplicateDeclaration' (class org.elm.ide.inspections.ElmDuplicateDeclarationInspection) was invoked for. Message: 'Multiple declarations with name 'rect''.
Element containing file: Elm File
Inspection invoked for file: Elm File
 [Plugin: org.elm.klazuka]
	at com.intellij.ide.plugins.PluginManagerCore.createPluginException(PluginManagerCore.java:405)
	at com.intellij.diagnostic.PluginProblemReporterImpl.createPluginExceptionByClass(PluginProblemReporterImpl.java:12)
	at com.intellij.diagnostic.PluginException.createByClass(PluginException.java:58)
	at com.intellij.diagnostic.PluginException.logPluginError(PluginException.java:75)
	at com.intellij.codeInsight.daemon.impl.LocalInspectionsPass.createHighlightsForDescriptor(LocalInspectionsPass.java:560)
	at com.intellij.codeInsight.daemon.impl.LocalInspectionsPass.addHighlightsFromResults(LocalInspectionsPass.java:501)
	at com.intellij.codeInsight.daemon.impl.LocalInspectionsPass.inspect(LocalInspectionsPass.java:207)
	at com.intellij.codeInsight.daemon.impl.LocalInspectionsPass.collectInformationWithProgress(LocalInspectionsPass.java:118)
	at com.intellij.codeInsight.daemon.impl.ProgressableTextEditorHighlightingPass.doCollectInformation(ProgressableTextEditorHighlightingPass.java:84)
	at com.intellij.codeHighlighting.TextEditorHighlightingPass.collectInformation(TextEditorHighlightingPass.java:55)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.lambda$null$1(PassExecutorService.java:429)
[...]

PR #646, opened 2020-04-09
Fixes for 'duplicate declaration' inspection

klazuka:
There were a couple of problems here.

1. Issue #631 reported an exception on the `master` branch caused by the 'duplicate declaration' inspection trying to report an error on a file _other than_ the file being inspected.
2. The inspection was also triggering false positives in certain shadowing contexts. Elm's shadowing rules are confusing and undocumented. It's apparently okay to shadow a name that you import from another module. So I relaxed the inspection in this case (which indirectly fixed the first problem/exception).
3. There was a bug in `ElmValueDeclaration.declaredNames()` where it erroneously included names from the right-hand-side of destructuring assignments. This was causing additional false positives.

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/psi/elements/ElmValueDeclaration.kt
+++ b/src/main/kotlin/org/elm/lang/core/psi/elements/ElmValueDeclaration.kt
@@ -79,7 +79,7 @@ class ElmValueDeclaration : ElmStubbedElement<ElmPlaceholderStub>, ElmDocTarget 
                 includeParameters -> assignee + assignee.namedParameters
                 else -> listOf(assignee)
             }
-            is ElmPattern -> descendantsOfType()
+            is ElmPattern -> assignee.descendantsOfType()
             is ElmOperatorDeclarationLeft -> when {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/elm/lang/core/psi/elements/ElmValueDeclaration.kt
```diff
@@ -79,7 +79,7 @@ class ElmValueDeclaration : ElmStubbedElement<ElmPlaceholderStub>, ElmDocTarget
                 includeParameters -> assignee + assignee.namedParameters
                 else -> listOf(assignee)
             }
-            is ElmPattern -> descendantsOfType()
+            is ElmPattern -> assignee.descendantsOfType()
```
ajalt, line 82: 🤦‍♂️

pravdomil:
more info about shadowing https://github.com/elm/compiler/blob/master/hints/shadowing.md

klazuka:
@pravdomil good find. I was unaware of whether this shadowing behavior was intentional or accidental.

Issue #613, opened 2020-02-04
"Add missing case branches" feedback


lydell:
I love this feature! But it could be tweaked a little.

### Use the same prefix as previous cases.

Input:

```elm
case foo of
    RemoteData.NotAsked ->
        x
```

Output:

```elm
case foo of
    RemoteData.NotAsked ->
        x

    NotAsked ->
        

    Loading ->
        

    Success a ->
```

Expected:

```elm
case foo of
    RemoteData.NotAsked ->
        x

    RemoteData.NotAsked ->
        

    RemoteData.Loading ->
        

    RemoteData.Success a ->
```

### Prefix based on import

Input:

```elm
import RemoteData exposing (RemoteData)


foo : RemoteData String String -> Int
foo remoteData =
    case remoteData of
```

Output:

```elm
import RemoteData exposing (RemoteData)


foo : RemoteData String String -> Int
foo remoteData =
    case remoteData of
        NotAsked ->
            

        Loading ->
            

        Failure e ->
            

        Success a ->
```

Expected:

```elm
import RemoteData exposing (RemoteData)


foo : RemoteData String String -> Int
foo remoteData =
    case remoteData of
        RemoteData.NotAsked ->
            

        RemoteData.Loading ->
            

        RemoteData.Failure e ->
            

        RemoteData.Success a ->
```

### Better indentation for inserted branches

Input:

```elm
foo : RemoteData String String -> List String
foo remoteData =
    [ "whatever", "foo" ]
        ++ (case remoteData of
                NotAsked ->
                    []
           )
```

Output:

```elm
foo : RemoteData String String -> List String
foo remoteData =
    [ "whatever", "foo" ]
        ++ (case remoteData of
                NotAsked ->
                    []


            Loading ->
                

            Failure e ->
                

            Success a ->
                           )
```

Expected (inserted patterns at same indent as previous, final closing parenthesis’ indentation not changed):

```elm
foo : RemoteData String String -> List String
foo remoteData =
    [ "whatever", "foo" ]
        ++ (case remoteData of
                NotAsked ->
                    []


                Loading ->
                    

                Failure e ->
                    

                Success a ->
           )
```

### Insert new branches in same order as type

Input:

```elm
type Msg
    = One
    | Two
    | NewCoolMsg
    | Three


update : Msg -> String
update msg =
    case msg of
        One ->
            "One"

        Two ->
            "Two"

        Three ->
            "Three"
```

Output:

```elm
type Msg
    = One
[...]

PR #648, opened 2020-04-10
Add missing branches now produces valid code

pravdomil:
as mentioned here https://github.com/klazuka/intellij-elm/issues/613

now it looks like this:
image

image


klazuka:
This was intentional so that the user would know that they have to fill it in. But maybe Debug.todo would be better. I'm going to close this out since I'd rather look at improvements to the entire feature.

PR #653, opened 2020-04-10
Incomplete pattern inspection applies at whole case expression

pravdomil:
if case expression is missing branches then whole case expression is invalid not only the keyword `case`

now it looks like this
image

### DIFF
--- a/src/main/kotlin/org/elm/ide/inspections/ElmIncompletePatternInspection.kt
+++ b/src/main/kotlin/org/elm/ide/inspections/ElmIncompletePatternInspection.kt
@@ -21,20 +21,20 @@ class ElmIncompletePatternInspection : ElmLocalInspection() {
             else -> arrayOf(AddWildcardBranchFix())
         }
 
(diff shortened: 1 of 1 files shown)


lydell:
I like that only `case` is underlined. Too much red is noisy. 🤷‍♂️

pravdomil:
can we do it so, only case is underlined, but the fix is available on whole case expression?

klazuka:
This is intentional (to avoid red squiggles everywhere).

pravdomil:
and what if only the case keyword is underlined, but the fix is available on whole case expression?

klazuka:
I’d rather that it corresponds with the squiggle.

пн, 17 авг. 2020 г. в 7:27 PM, pravdomil <notifications@github.com>:

lydell:
I would like the squigglies to be on the `of` keyword, like Elm for VSCode, as mentioned at the end of https://github.com/klazuka/intellij-elm/issues/613. Then you don’t have to jump back to the beginning of the line to autofill the branches.

PR #670, opened 2020-06-29
Pipeline

dillonkearns:
Hello @klazuka! I've got this to/from pipeline intention working pretty robustly. I'm happy to make any changes that you can think of, I'd love to hear your thoughts and ideas. I'm also happy to add more test coverage if you can think of areas that I haven't covered in the test suite.

Here are some demos with the code set to run `elm-format` after invoking each operation.

## Questions
* Should I run `elm-format` at all? Right now I have it set to run after running the pipeline intention if  `isElmFormatOnSaveEnabled` is true.
* Is this better split into multiple intentions to give it more granular naming?

### Examples 1 & 2
Screen Recording 2020-06-29 at 02 04 PM

### Examples 3 & 4
Screen Recording 2020-06-29 at 02 06 PM

### Example 5
Screen Recording 2020-06-29 at 02 07 PM


dillonkearns:
A couple of additional thoughts on things to work on to get this ready to merge:

- The whitespace handling in the current PR is hacky and relies on elm-format. Skip running elm-format there, and instead use some logic similar to https://github.com/klazuka/intellij-elm/blob/master/src/main/kotlin/org/elm/lang/core/psi/ElmPsiFactory.kt#L147-L171
- Add some test cases for examples that mix <| and |>
- Add some handling for converting chains of <|

klazuka:
Cool.

I tested it out on elm-spa-example `src/Article.elm`and ran into some cases where it generates code that doesn't type-check. For instance,

```elm
fetch : Maybe Cred -> Slug -> Http.Request (Article Full)
fetch maybeCred articleSlug =
    Decode.field "article" (fullDecoder maybeCred)
        |> Api.get (Endpoint.article articleSlug) maybeCred
```

And also if you play around with un-pipelining the functions that use json-decode-pipeline, it also generates invalid code. I think the problem is that it doesn't properly handle the introduction (or removal) of parentheses.

For instance:

```elm
foo = 
    identity "bar"
        |> identity
```

If you try to un-pipeline that, I would expect `identity (identity "bar")`, but it gives you something unexpected.

Similarly, if you try to pipepline `identity (identity "bar")`, it gives `"bar" |> identity |> identity`, but I would expect `identity "bar" |> identity`.

dillonkearns:
Great idea to look to `elm-spa-example` for test cases. And thanks for finding those unhandled scenarios, I'll add some tests for those and see if I can find any others. Thanks!

klazuka:
As for splitting the intentions, yes, I think that's a good idea. One problem with it that I have right now is that the "pipeline" action is suggested *everywhere*. And in some cases it's hard to understand what it will do.

One improvement would be to have a separate "un-pipeline" intention that is only offered on the `|>` operator. That will help make things more clear.

As for how and when to _create_ a pipeline, I think we need to enumerate the cases where the action is useful to see if we can come up with some rules for when it should be made available. This will help with the implementation and hopefully make it easier for the user to reason about when the action is available and what it will do.

klazuka:
As for relying on elm-format to do the formatting, that's a good idea, although I'm not sure about the performance ramifications. You definitely don't want to synchronously run an external process while executing inside IntelliJ's read or write action. I haven't looked at how you're doing it, but the IntelliJ docs on [their threading rules](https://www.jetbrains.org/intellij/sdk/docs/basics/architectural_overview/general_threading_rules.html) is worth reading.

klazuka:
Here's an example that generates invalid code when removing pipes, but *only* if the caret is on the right-hand-side of the pipe operator:

```elm
f =
    List.range 0 5
        |> List.map identity
```
Try it with the cursor on `List.range` (works correctly) and with the cursor on `List.map` (fails to type-check).

klazuka:
It seems to me that the 'remove pipes' intention should only be available when the cursor is actually on the `|>` token. Is there a reason you made it more lenient?

klazuka:
As for the 'add pipes' intention, it's hard for me to predict what it's going to do. Take this example:
```elm
foo =
    List.length
        [ List.range 0 5
            |> List.maximum
        ]
```
Since you can basically invoke "add pipes" anywhere, what should the user's mental model be of the _scope_ of the transformation? For instance, if I'm inside the List literal, is it scoped to a single element of that List? That sounds reasonable to me. I think it'd be helpful if there was some sort of rule like "parens, square brackets and curly brackets delimit the scope of the transform." 

I don't know what the actual rule _should be_, but nailing that down would help both implementation _and_ user understanding.

klazuka:
Also, I noticed a bug with the following example:
```elm
foobar =
    identity [ negate 1 ]
```
if you put the cursor on `identity` (or even at the end, after the closing square bracket) and invoke 'add pipeline', the generated code strips out the List literal. The code will type-check, but now the type of the expression is `number` instead of `List number`.

dillonkearns:
Hey Keith, thanks so much for taking a look at the progress!

I just pushed some new code where this example is fixed:

Screen Recording 2020-07-04 at 09 05 25 AM

> It seems to me that the 'remove pipes' intention should only be available when the cursor is actually on the |> token. Is there a reason you made it more lenient?

My thought process for that was that if you're in the context of a pipeline, it's convenient to be able to just pull up that action without moving your cursor to that precise operator. At least for me as a user, I would find that more convenient. So for the logic on whether the remove intention is available, I simply check if the function call in the current level of scope has any pipes before or after as siblings (i.e. it's in a pipeline). What do you think of that as the context for showing the intention?

> ```
> foobar =
>     identity [ negate 1 ]
> ```

Thanks for finding this case, I'll add a test for this!

> Since you can basically invoke "add pipes" anywhere, what should the user's mental model be of the scope of the transformation? For instance, if I'm inside the List literal, is it scoped to a single element of that List? That sounds reasonable to me. I think it'd be helpful if there was some sort of rule like "parens, square brackets and curly brackets delimit the scope of the transform."

Yes, that's exactly what I had in mind, too! There are some kinks to it right now that I need to work out, but that's the intention, to find the containing scope and apply it there. So the pipelines will be added/removed from the siblings of the function call you're looking at.

One interesting case is if you have pipelines already, but don't have the entire thing in a pipeline normalized form, I think it would be helpful to still make the Add Pipeline intention available to turn it into a full pipeline chain. For example:

```elm
example =
    times2 5
        |> List.singleton
```
*Add pipeline |>*

```elm
example =
    5
        |> times2 
        |> List.singleton
```

I don't have it working that way at the moment, but that's the direction I'm thinking of.

dillonkearns:
An internal note about the code: I'm thinking that it will be a generally helpful function to have available to get the function being called and all of its arguments in a normalized form. For example, it could be defined in the `ElmFunctionCallExpr` class. I think this could be helpful for some other refactorings that I'm thinking about working on.

There isn't anything to help normalize the form independent of pipes now, right? I think that could be a nice core building block to have available to ask, for a given function call expression, what values are being passed in.

klazuka:
Your reasoning about when to make 'remove pipes' available is fine with me.

> There isn't anything to help normalize the form independent of pipes now, right?

Do you mean normalizing expressions involving `|>` and `<|` into something just using parenthesized expressions? For instance,
```elm
f <| (g |> h)
```
becomes
```elm
f (h g)
```
?

If so, no, we don't have anything like that right now. It sounds useful, but the variety of whitespace issues related to this sounds complicated. Sometimes it will be multiple lines, sometimes parenthesized, etc. etc. I would worry that if you start to work on this, it could make an already difficult task more difficult (although certainly useful in the long-run).

dillonkearns:
Hey Keith,

I've got the pipeline intention showing up when it's relevant, operating on the desired context, and resulting in valid transformations. I'm not aware of any remaining cases that aren't working correctly besides handling of whitespace.

So what's remaining on my todo list is:

- [ ] Handle whitespace correctly for cases that are whitespace-sensitive
- [ ] Support `<|` chains in Remove Pipeline intention
- [ ] The intention is depending on `elm-format` to clean things up after running. Maybe I should not show the intention if `elm-format` isn't set up in the toolchain?


Also, I'm not able to reproduce the build failure. Do you have any idea what the problem might be or how I could reproduce it?

```
org.elm.workspace.ElmWorkspaceResolveTest > test does not resolve modules which are not direct dependencies FAILED
    java.lang.IllegalArgumentException at ElmWorkspaceResolveTest.kt:455

1313 tests completed, 1 failed

> Task :test FAILED

Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':test'.
> There were failing tests. See the report at: file:///Users/runner/work/intellij-elm/intellij-elm/build/reports/tests/test/index.html
```

That's all I've got on my todo list right now. Let me know if you can think of anything else.

klazuka:
I'll look at your PR now.

As for that test failure, that has been periodically failing on Mac and Linux for a while now. It doesn't have anything to do with your PR. You can ignore it.

dillonkearns:
Oh, good to know about the test failures.

Thanks for taking a look. By the way, I played around with 3 different approaches to whitespace formatting:

1) Always run `elm-format` (current option)
2) Run the new element being created in the PsiFactory through `elm-format`
3) Don't run `elm-format` (just try to format reasonably)

(2) seems like an improvement over (1). It doesn't handle removing unnecessary parens around the main expression, so that's the main downside. The performance seems comparable, though. I'm guessing that there's a lot of overhead to just running the external executable.

(3) is significantly faster. It feels way snappier. I've got a local change where I am able to get the whitespace looking good (using some logic similar to your extract to let).

I'm actually learning towards (3). The biggest remaining challenge there will be avoiding unnecessary parens (but added them when needed). `elm-format` does a great job getting rid of those, so right now I put parens around each part of the pipeline.

If I can only put it around expressions where it's needed, then that would be ideal, as long as there aren't a lot of corner cases. For example, lambdas need to get parens to keep the same precedence. I'll have to do some experimentation to see what other cases need it and if it's feasible to manually handle all those cases.

dillonkearns:
Oh yes, and there's now a `Remove Pipeline` intention available for `<|`s

image

image

klazuka:
Of the 3 options for formatting, "don't run elm-format" would be ideal. Invoking that external process, especially through npm, has been problematic. I read that some improvements have been made in elm-format's wrapper script, but if we could avoid it, that'd be great.

But if handling the formatting yourself ends up being a ton of corner-cases, then just stick with the elm-format approach.

klazuka:
So I've played around with your latest code a bunch, and it works great! 💯 It feels *way* more bullet-proof. It's really cool that you can quickly preview whether the code would be more/less readable using a pipeline.

The only problem I've noticed is that it doesn't preserve comments. Here's an example from elm-spa-example:
```elm
scrollToTop =
    Dom.setViewport 0 0
        -- It's not worth showing the user anything special if scrolling fails.
        -- If anything, we'd log this to an error recording service.
        |> Task.onError (\_ -> Task.succeed ())
```

If you do 'remove pipes' on that, the line comments disappear. Same thing happens if it's a block comment. I'd be ok with deferring this to a later PR though as the fix may be hard and the workaround is pretty obvious (undo, copy the comment, run the intention again, paste the comment somewhere).

dillonkearns:
Hey Keith! Thanks so much for taking another look.

Quick update. I was able to get things formatting nicely directly (without `elm-format`). It feels way snappier, and I've got it to a state where the results are looking similar or exactly the same as `elm-format`'s results in most cases.

After working on it a bit, I realized that the only times that parens are needed within a pipeline are for syntax that changes precedence. And the only syntax that changes precedence is parens or operators (i.e. ElmBinOpExpr). So it turns out there aren't too many corner cases to consider!

Regarding comments, I handle one or two cases for that already, but in most cases, as you pointed out, I'm throwing comments away. But I do at least have some code that shows it's possible to preserve comments in a reasonable way. It will take some work to get it working, for example, to preserve comments in a reasonable way in the `Remove Pipeline` intention. I'll have to format each part of the pipeline on a separate line. It will be a bit tedious, but doable. If it's too much trouble, I'll just hide the intention in that case, but it would be great to get that last piece working so that people can use these refactorings without any hesitation.

I'll keep you posted! After that, I think it will be ready to merge in after a final round of testing and code cleanup.

Thanks again for the help with this!

klazuka:
Great! I’ll look at the code closely once you are ready for review, and do
a final round of play-testing.

вс, 12 июля 2020 г. в 10:11 PM, Dillon Kearns <notifications@github.com>:

dillonkearns:
Hey @klazuka, I'm now preserving comments. There's some code cleanup work I'd like to do before merging this in, but in terms of the behavior I think this is a candidate for the release. Could you take a look and let me know your thoughts? If it looks good then I'll finish the code cleanup work and we can finalize this.

A couple of things to note:

- Often we can simply do a pipeline where each pipeline segment is on its own line. In these cases the formatting is quite clean.
-  If there are any comments or multi-line arguments then I opt for using some extra parens to be sure I preserve the comments and whitespace correctly. It's very hard to preserve the comments otherwise, but most of the time that's not an issue so this seems reasonable. And elm-format will fix it when you save.

I'm open to any and all feedback and would love to hear any thoughts you have! Thank you!

klazuka:
I tested it out with jxxcarlson/elm-tar which has a lot of pipeline code. I only found one problem, and it's a pretty minor edge case.

If you try to remove the pipeline from:
```elm
foo =
    [ 1
    , 2
    , 3
    ]
        |> identity
```
it will generate a bad ident like so:
```elm
foo =
    identity
        [ 1
    , 2
    , 3
    ]
```

This a pretty minor issue because, at least in this case, it is still syntactically valid and running elm-format will fix it. So if it's hard to fix, then I'd be ok with leaving it as a known issue.

I think it's ready for final review. Can you go through and delete any commented-out code? Plus whatever other code cleanup, and then let me know and I will look more closely at the implementation.

dillonkearns:
Hi Keith, thank you for testing it out!

Yeah, let's leave that case as a known issue for now. That was one of the multiline cases that I got to "not working ideally, but working correctly." I did a lot of work to try to refine those cases, but didn't quite iron out that case. But I think it will overall be a smooth experience with the current behavior 👍 

I'll do some final code cleanup and reply here when it's ready for final review. Thanks!

dillonkearns:
Hey @klazuka, I did some refactoring and this PR is now ready for review. Let me know what you think, I'm more than happy to address any suggestions you have. Thank you!

klazuka:
Nice! I played around with a lot and it worked really well. Test coverage looks good and the overall code was structured well. There were a few code style things, but I prefer to just do the cleanup myself rather than try to describe the changes in PR comments, especially since some code style issues I really need to play around with to decide what fits best.

PR #674, opened 2020-08-02
Add parameter inlays

dillonkearns:
Hello! I was playing around with an inlay parameter hint provider, inspired by the [`intellij-rust`](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/hints/parameter/RsInlayParameterHintsProvider.kt) implementation.

I wanted to get an initial version to you for feedback, along with some notes of the current behavior and some future possibilities. It turns out, this all worked out quite smoothly, although there are some interesting cases that I note below. Overall, I think this is pretty interesting with the current cases implemented as is, and we can iterate and refine from there.

I have a few todos that I'd like to take care of before merging that I've written out below as well. If you think of anything else, let me know and I'll add it to the list.

### Pipelines
Right now, I'm now showing hints for pipelines. This is a good thing in general, because usually the argument that you're piping through is understood to be the subject of the pipeline, so it doesn't need a name.

Image 2020-08-01 at 6 38 42 PM

#### Destructuring a Custom Type with a single constructor value

Shows a hint with the destructured name.

Image 2020-08-01 at 6 34 10 PM

#### Destructuring records
I believe the best way to handle these cases would be to not show a hint at all for these arguments:
```elm
example =
    nameToString
        { first = "Jane"
        , last = "Doe"
        }


nameToString { first, last } =
    first ++ " " ++ last

```

This is on my todo list, because right now it's using one of the destructured names, but that's not accurate (whether there is only one value destructured, or more than one as in the above example).

#### Underlying kernel function
It turns out, quite a few of the standard library definitions use this point free style. Like `Decode.map`, for example.

```elm
map : (a -> value) -> Decoder a -> Decoder value
map =
    Elm.Kernel.Json.map1
```

For cases like this, I don't think there's much you can do since it's defined in the kernel.

#### Non-Kernel point-free definitions
Longer term, if a function is point-free, but its underlying definition is non-kernel code, we can look at the argument name one level deeper. For example:

```elm
example =
    heading "This is content" "This is postfix"
            -- ^ content
                                       -- ^ postfix

heading =
    combine "Heading:"

combine prefix content postfix =
    prefix ++ "\n" ++ content ++ "\n" ++ postfix
```

We can infer those values that I annotated. I think it makes sense to defer this to a future pull request, after we see how it goes with the first set of functionality, though.

## Remaining TODOs

- [x] Don't show hints for destructured records (including 1 argument, don't show for any of these cases)
- [ ] Hide hints for arguments that only have a single argument
- [ ] Don't show hints for arguments that end with the passed in variable name
- [x] Don't show hints for parameters named `_`

#### Pipeline type names
Show the types along the way in the pipeline. Sometimes it's hard to keep track of the intermediary types, and there's prior art. See https://www.jetbrains.com/help/rider/Inline_Parameter_Name_Hints.html#type-name-hints

image

#### Applicative pipeline field names
I think it could solve a real problem that comes up a lot that it's hard to tell which decoder value will map to which record value because of the positional nature of pipelines. [This recent Discourse thread](https://discourse.elm-lang.org/t/about-the-ergonomics-of-applicative-json-decoding/6099/14) discusses it, and there have been many similar discussions in the past.

#### Lambda type hints
As with the Rider IDE, we could display type hints for lambdas. I often get confused when I'm writing a `foldl` expression, for example, and can't keep the order of the arguments straight. Having the type names would help for this kind of case.

Here's what it looks like in Rider (see https://www.jetbrains.com/help/rider/Inline_Parameter_Name_Hints.html#type-name-hints)

image

### DIFF
--- a/src/main/kotlin/org/elm/ide/hints/parameter/ElmInlayParameterHints.kt
+++ b/src/main/kotlin/org/elm/ide/hints/parameter/ElmInlayParameterHints.kt
@@ -10,6 +10,8 @@ import com.intellij.codeInsight.hints.Option
 import com.intellij.psi.PsiElement
 import org.elm.lang.core.psi.*
 import org.elm.lang.core.psi.elements.*
+import org.elm.lang.core.types.findTy
+import org.elm.lang.core.types.renderedText
 
 @Suppress("UnstableApiUsage")
 object ElmInlayParameterHints {
@@ -25,6 +27,11 @@ object ElmInlayParameterHints {
 
     @ExperimentalStdlibApi
     fun provideHints(elem: PsiElement): List<InlayInfo> {
+        if (elem is ElmAnonymousFunctionExpr) {
+            return elem.namedParameters.map { param ->
+                InlayInfo(": " + param.findTy()?.renderedText(), param.endOffset)
+            }
+        }
         val (callInfo, valueArgumentList) = when (elem) {
             is ElmFunctionCallExpr -> ( elem.target.reference?.resolve() to elem)
             else -> return emptyList()
--- a/src/main/kotlin/org/elm/ide/hints/parameter/ElmInlayParameterHints.kt
+++ b/src/main/kotlin/org/elm/ide/hints/parameter/ElmInlayParameterHints.kt
@@ -45,7 +45,6 @@ object ElmInlayParameterHints {
                         return@map InlayInfo("$asText:", arg.startOffset)
                     }
                     when (param) {
-                        is ElmRecordPattern -> null
                         is ElmAnythingPattern -> null
                         is ElmPattern -> {
                             val child = param.unwrapped
--- a/src/main/kotlin/org/elm/ide/hints/parameter/ElmInlayParameterHints.kt
+++ b/src/main/kotlin/org/elm/ide/hints/parameter/ElmInlayParameterHints.kt
@@ -64,6 +64,9 @@ object ElmInlayParameterHints {
                                 emptyList()
                             }
                         }
+                        is ElmUnitExpr -> {
+                            emptyList()
+                        }
                         is ElmUnionPattern -> {
 //                                listOf(InlayInfo(unwrapped.upperCaseQID.fullName + ":", arg.startOffset))
                             emptyList()
--- a/src/main/kotlin/org/elm/ide/hints/parameter/ElmInlayParameterHints.kt
+++ b/src/main/kotlin/org/elm/ide/hints/parameter/ElmInlayParameterHints.kt
@@ -39,13 +39,27 @@ object ElmInlayParameterHints {
         val hints = elements.zip(valueArgumentList.arguments.toList())
 
         return hints
-                .filter { (param, arg) ->
+                .map { (param, arg) ->
                     when (param) {
-                        is ElmRecordPattern -> false
-                        is ElmAnythingPattern -> false
-                        else -> true
+                        is ElmRecordPattern -> null
+                        is ElmAnythingPattern -> null
+                        is ElmPattern -> {
+                            val child = param.child
+                            if (child is ElmUnionPattern) {
+                                if (child.argumentPatterns.count() == 1) {
+                                    InlayInfo(child.argumentPatterns.first().text + ":", arg.startOffset)
+                                } else {
+                                    null
+                                }
+                            } else {
+                                InlayInfo(param.text + ":", arg.startOffset)
+                            }
+                        }
+                        else -> {
+                            InlayInfo(param.text + ":", arg.startOffset)
+                        }
                     }
                 }
-                .map { (hint, arg) -> InlayInfo(hint.text + ":", arg.startOffset) }
+                .filterNotNull()
     }
 }
--- a/src/main/kotlin/org/elm/ide/hints/parameter/ElmInlayParameterHints.kt
+++ b/src/main/kotlin/org/elm/ide/hints/parameter/ElmInlayParameterHints.kt
@@ -8,8 +8,7 @@ package org.elm.ide.hints.parameter
 import com.intellij.codeInsight.hints.InlayInfo
 import com.intellij.codeInsight.hints.Option
 import com.intellij.psi.PsiElement
-import org.elm.lang.core.psi.elements.ElmFunctionCallExpr
-import org.elm.lang.core.psi.elements.ElmFunctionDeclarationLeft
+import org.elm.lang.core.psi.elements.*
 import org.elm.lang.core.psi.startOffset
 
 @Suppress("UnstableApiUsage")
@@ -33,14 +32,20 @@ object ElmInlayParameterHints {
         if (callInfo == null) return emptyList()
         val elements =
                 if (callInfo is ElmFunctionDeclarationLeft) {
-                    callInfo.namedParameters.map {
-                        it.name
-                    }.toList()
+                    callInfo.patterns.toList()
                 } else {
                     emptyList()
                 }
         val hints = elements.zip(valueArgumentList.arguments.toList())
 
-        return hints.map { (hint, arg) -> InlayInfo("$hint:", arg.startOffset) }
+        return hints
+                .filter { (param, arg) ->
+                    when (param) {
+                        is ElmRecordPattern -> false
+                        is ElmAnythingPattern -> false
+                        else -> true
+                    }
+                }
+                .map { (hint, arg) -> InlayInfo(hint.text + ":", arg.startOffset) }
     }
 }
--- a/src/main/kotlin/org/elm/ide/hints/parameter/ElmInlayParameterHints.kt
+++ b/src/main/kotlin/org/elm/ide/hints/parameter/ElmInlayParameterHints.kt
@@ -39,23 +39,35 @@ object ElmInlayParameterHints {
         val hints = elements.zip(valueArgumentList.arguments.toList())
 
         return hints
-                .map { (param, arg) ->
+                .flatMap { (param, arg) ->
                     val asText = ( param as? ElmPattern )?.patternAs?.text
                     if (asText != null) {
-                        return@map InlayInfo("$asText:", arg.startOffset)
+                        return@flatMap listOf(InlayInfo("$asText:", arg.startOffset))
                     }
                     when (param) {
-                        is ElmAnythingPattern -> null
+                        is ElmAnythingPattern -> emptyList()
+                        is ElmTuplePattern -> {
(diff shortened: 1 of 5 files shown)


dillonkearns:
Here are some more cases as they are currently handled. One thing to note, I'm using top-level `as` clauses for names, which I think works well.

The `(_,y)` hint in `example2` also seems readable and useful, but I'd be interested to hear your opinion on that. I'm definitely open to not showing hints in those cases if that doesn't seem helpful.

Image 2020-08-02 at 8 46 38 AM

dillonkearns:
I made a few changes from the last update:

- I was able to show the hints for those destructured tuples inline
- I decided to skip record destructure hints for now since they can get noisy and can contain low-level details
- I am skipping hints for Custom Type destructuring for now since those hints don't seem to be useful too often from looking through some examples with that on
- I'm still using `as` clauses whenever they are provided

Image 2020-08-02 at 9 29 10 PM


When you pass in a non-literal tuple, I currently don't provide a hint for that. I could easily show a hint like `( x, y )` here:

```elm
example coordinates =
    xPosition coordinates

xPosition ( x, y ) =
    x
```

I'm thinking it's reasonable to skip that for now, though, and revisit that later.

## Summary

I think this is generally in pretty good shape, besides those two remaining todos. I'm open to any and all feedback, so feel free to share any constructive criticism or concerns you might have with any of the hints.

I was having some trouble getting the setting to save as well, I'm not sure if that's an issue with the `runIde` gradle task, or if I'm doing something wrong. I tried to follow the `intellij-rust` example there, but it doesn't seem to persist when I toggle the Elm-specific setting off for `SHOW_PARAMETER_HINT`. I'll work on that more as well, and it's good to know that we can add nuanced options in case users want to configure specific behavior (the Kotlin hints seem to have several options, for example).

klazuka:
I'll give it a try in the next few days. I can see how this would be useful, and I trust that your GraphQL examples are a good motivating factor. Personally, I turn off inlay hints for Kotlin and Java.

I have mixed feelings about showing parameter names as part of the API of a function in Elm. In most languages it's totally normal, but in Elm they are really more an internal name than an outward-facing name. For instance, if you look at the [docs for elm/core List](https://package.elm-lang.org/packages/elm/core/latest/List), the function signature only shows the type of parameters, never their name.

As you mentioned, point-free style is problematic since there are no names for the parameters. Destructuring is also problematic. And even in the case where the argument is named, the habit is to choose a name that makes sense internally, but not necessarily externally.

In any case, I will give your code a try and see how it feels.

klazuka:
One challenge in this is going to be the “stubs” system. For features like this that run automatically on multiple elements in the editor and which depend on analysis of imported files, the Psi queries should be able to get the data directly from the stubs. Otherwise, what will happen is that you will end up reading in all imported files and parsing them, which gets slow.

Our stub index includes the name of a function but not its parameters (see `ElmFunctionDeclarationLeftStub`). We also include the type annotation in the stub index.

We *could* add parameters / destructuring-pattern to the stub index, but that will increase the size of the index. And it won’t help with some of the fancier dereferencing that you wanted to do (re: point-free style) since you would need to reach into the function body to chase down a named binding.

I think the only path forward is to try adding function parameters to the stub index and see how much it increases the size of the index, RAM usage, project load time for a real-world project. Unfortunately I don’t have any existing technique for measuring those things, so it would need some investigation.

dillonkearns:
Hey Keith, thanks so much for taking a look and sharing that info.

That's a great point about the stub indexes. Regarding looking through point-free definitions and finding the underlying definition, those performance considerations sound like more reason not to pursue that functionality, at least for now.

It also seems like pattern matching info is probably not worth including in the indexed function data. In some cases, the names may not be high-level, so it seems like it's not worth the effort and potential performance cost to include that extra metadata.

From playing around with it, it seems that the most useful names are the regular parameter names. I'll play around with some performance profiling. It looks like there are some tools for that here: https://intellij-support.jetbrains.com/hc/en-us/articles/207241225-Performance-testing-plugin. I'll let you know what I find. I'm not sure what it would entail to do add parameter names to the stubs, but I'll take a look at that and think about the tradeoffs and effort there.

## Type Hints

What I think will be most valuable by far are type hints.

For example, type hints for lambdas:

Image 2020-08-03 at 2 43 59 PM

Since lambda parameters don't have type annotations, I find myself using ctrl-shift-p often to get context on what data is available.

I also have an initial version working for providing type info for let bindings that aren't annotated, and I think this can be useful. And I'm also working on adding annotations for intermediary types in a pipe chain.

Adding type hints calls to `ElmPsiElement.findTy()`. Do you think that will use information that will be indexed already since it's just resolving types? Here's the basic implementation. It doesn't handle all the cases, like excluding annotations when the type is unknown. But it already works pretty well:

https://github.com/dillonkearns/intellij-elm/blob/ea8ac76d3b7190ead672692d494ed4f0ac04315e/src/main/kotlin/org/elm/ide/hints/parameter/ElmInlayParameterHints.kt#L32-L41

## Pipeline Type Annotations
For the pipeline type annotations, I'll need to expose some type information that's being computed here:

https://github.com/dillonkearns/intellij-elm/blob/977792141195be51ab9da901885024ae489a576c/src/main/kotlin/org/elm/lang/core/types/TypeInference.kt#L345-L346

That will give me access to the type of each piece of the pipe chain. Since it's already being computed here, I'll just need to include that data in that `Ty` data type. Does that seem like a reasonable idea?

Thanks again for taking the time to look at this. I'm happy to do a call some time if it's easier to discuss more details that way.

klazuka:
> Adding type hints calls to ElmPsiElement.findTy(). Do you think that will use information that will be indexed already since it's just resolving types?

Yes, the type inference system is already designed to work strictly in terms of stub data so that it never needs to parse any files other than the one you are editing.

> For the pipeline type annotations, I'll need to expose some type information ... That will give me access to the type of each piece of the pipe chain. Since it's already being computed here, I'll just need to include that data in that `Ty` data type. Does that seem like a reasonable idea?

The `Ty` values are heavily cached, so any changes to that data will need some extra scrutiny by either AJ or me. But I'm wondering if there's a better way. It sounds like you need more fine-grained sub-expression type inference. This is tracked in `InferenceResult.expressionTypes`, and it is how the `ctrl-shift-p` (lookup type of expression) feature works (see `ElmExpressionTypeProvider`). I'm not entirely sure what you want to do, but it may be that we just need to break apart the pipeline expression and register it in `expressionTypes`. This will result in higher memory usage, but AJ did a lot of work earlier this year to make the cached type inference data use less memory. In any case, let's keep type inference changes out of this PR so that we can look at a smaller change later with AJ.

klazuka:
In terms of performance, I guess the main thing I'm curious about is the increase to the size of the stub index on disk/memory, whichever is easier to calculate. Unfortunately I don't know off-hand how to investigate that.

But I do have a couple tricks to detect when code accidentally parses a file when it should instead be getting the data from the stub index.
1. for unit tests, see `ElmResolveTestBase.stubOnlyResolve()`, specifically `checkAstNotLoaded`
2. for running the IDE, set a breakpoint on `PsiFileImpl.loadTreeElement()`line 196. Whenever the breakpoint hits, log the name of the file and continue. Then when you are running the IDE, just keep an eye on the log output and make sure that it only prints out the name of the currently edited file.

image

The only time when the IDE should go out and parse a bunch of files is when the user explicitly invokes an expensive task like `Find Usages`. It should never parse external files when running an inspection or any other kind of passive annotator.

dillonkearns:
> Yes, the type inference system is already designed to work strictly in terms of stub data so that it never needs to parse any files other than the one you are editing.

That's great news! Maybe to keep things simple, I can focus on having this first pull request introduce some hints for types like lambda types and unannotated let bindings. Then I can break up other features into other PRs so it's more manageable to evaluate any performance implications.

## Pipeline Hints

I can also do the pipeline type hints as separate PR. To give you a sense of what the pipeline type inference might look like, here's a very simple implementation. I'm not sure if this plays well with the cached types, probably not given that it introduces a new field.

I guess instead of introducing a new field, I could just use the existing `expressionTypes` field and then it would get cached? It sounds like that would make it easier in terms of relying on existing caching logic instead of introducing new data that needs caching.

https://github.com/klazuka/intellij-elm/commit/4716f24aabed6ab8810c25001b9bdeefec6b84a4

And here are a few examples of the above implementation in action:

Image 2020-08-05 at 9 13 54 AM


Image 2020-08-05 at 7 48 37 AM

Image 2020-08-04 at 10 25 56 PM

dillonkearns:
If we were to use the existing `expressionTypes` to cache the computed types, it would be pretty straightforward I think as long as we can build up those expressions.

Is it valid to build up an expression that's only part of an `ElmBinOpExpr`, though? That might be the sticking point to that approach. Take this example:

```elm
123.45
    |> round
    |> List.singleton
```

We would need to store the expressionType for `123.45` (not a problem since that's already an expression).

And `123.45 |> round`. But I'm not sure if we can do that? This code builds it up into its own data structure to evaluate it and compute the precedence and types of the sub-expression. But can we build an `ElmPsiElement` out of it that we can then use to index into the hash? Does it get the value out of the hash by reference equality check, or something deeper?

https://github.com/dillonkearns/intellij-elm/blob/977792141195be51ab9da901885024ae489a576c/src/main/kotlin/org/elm/lang/core/types/TypeInference.kt#L358

klazuka:
It may be that the reason AJ doesn't register the types of the sub-expressions within an infix operator expr is due to the representation problems. As you've noticed, he had to build a temporary parse tree because precedence is not known during the normal parser pass.

I don't think we will be able to cook-up an `ElmPsiElement` because these are stateful things that are deeply hooked into IntelliJ magic. But I'm certain that there's a way to make this work. For now, let's keep the focus on low-hanging fruit and then deal with extending type inference data later.

dillonkearns:
> I don't think we will be able to cook-up an ElmPsiElement because these are stateful things that are deeply hooked into IntelliJ magic.

That's what I was thinking, too.

> For now, let's keep the focus on low-hanging fruit and then deal with extending type inference data later.

I totally agree 👍 I'll focus on getting the lowing hanging fruit to a released state, and then revisit these things that require larger architectural changes after that. I'll work on getting this PR ready with a focus only on type hints (and not for intermediary pipeline values).

cies:
Hi @dillonkearns ! Are you still working on this, or can I close this issue. (maybe you can close it yourself explaining where you stalled)

dillonkearns:
Hi @cies, the reason I didn't merge this was because Keith had some concerns about performance. I'll close this, but feel free to make use of it if you'd like to.

PR #678, opened 2020-08-08
List.map to List.Foldr refactoring

dillonkearns:
Hey @klazuka, here's a PR for the `List.map` -> `List.foldr` refactoring that @avh4 and I built in the live stream.

Here's a demo:

Screen Recording 2020-08-08 at 03 20 42 PM

A few notable features:

- If the foldr argument is multiline, it will do multiformatting formatting. Otherwise it will format on a single line.
- It preserves whitespace to ensure that the semantics of any `case` statements or `let` bindings are preserved (and so it feels smooth and doesn't reformat twice)
- If the new lambda argument names are in scope (`result` and `item`), it will add a number after the name until it's unique in that scope
- The refactoring doesn't present itself if there are 0 arguments passed in to `List.map`
- The refactoring works with 1 or 2 arguments (i.e. partially applied or not)

I considered doing something to do a live rename, like rename refactorings, but it seems unnecessary since you can rename it using that refactoring if you'd like to. Overall, it's feeling like a nice little win!

I'm open to any feedback you have on the functionality or the code. This one is a lot simpler than the other two I've taken on, so hopefully this one will be easier to wrap up.

### DIFF
--- a/src/main/kotlin/org/elm/ide/intentions/MapToFoldIntention.kt
+++ b/src/main/kotlin/org/elm/ide/intentions/MapToFoldIntention.kt
@@ -1,0 +1,41 @@
+package org.elm.ide.intentions
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import org.elm.lang.core.psi.ElmPsiFactory
+import org.elm.lang.core.psi.ancestors
+import org.elm.lang.core.psi.elements.ElmExposingList
+import org.elm.lang.core.psi.elements.ElmFunctionCallExpr
+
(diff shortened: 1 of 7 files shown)


klazuka:
Looks good. I tried it out on various examples, and it worked really well. Nice job.

dillonkearns:
Thanks for testing it out Keith! More to come. I'm making progress on the final touches of the pipeline refactoring right now.

klazuka:
FYI there is a big change (d37b8ed) on the master branch that touched a lot of files and changed the stub structure. This PR merged cleanly, but you should probably rebase the other ones. When running the IDE after rebasing, you may need to do `File -> Invalidate Caches... -> Invalidate and Exit', then run the IDE again (because the stub structure changed; normal installs via the plugin repository don't require this step because IntelliJ will automatically re-index).

PR #679, opened 2020-08-09
Cleanup MapToFoldIntention

klazuka:
@dillonkearns I went ahead and made your code more closely match our Kotlin style.

The multi-line, indented code-gen is definitely a pain. I really regret not doing the work to figure out how to integrate `elm-format` into IntelliJ's whitespace formatting system.


dillonkearns:
It's good to see the diff here, thanks for sharing 👍

I'm hoping that over time we'll build up some nice abstractions for dealing with formatting and refactoring. The unique name as a helper class here looks like a great step in that direction.

klazuka:
Yep. I think this `IndentedTextBuilder` can be useful more widely when generating code. I have a branch experimenting with extending it to the "introduce variable" refactoring (let/in extraction). It's still a bit awkward, but I think it's going to be an improvement over the old, ad-hoc way.

Issue #680, opened 2020-08-14
Autocomplete Record Field names


ryancerf:
Before completion.
```elm
type alias Foo = { name : String }

f : Foo
f =
    { n{-caret-} }
```

After completion
```elm
type alias Foo = { name : String }

f : Foo
f =
    { name = {-caret-} }
```

ryancerf:
Hi,

First off, thanks for the awesome plugin. I started playing around with a few possible implementations and would love to get your advice on how to approach this.

Couple of options:

1) Use the existing diagnostics field on `InferenceResult`.
Parse the diagnostics on `InferenceResult` looking for a `TypeMismatchError` with missing fields on the `recordDiff`. This is easy, but will not work in some situations such as nested records. It will probably get us 60% of the way there and I could send a PR in a day or two.

2) Add information to `InferenceResult` that identifies the expected (declared) types of expressions.
This is a much heavier lift since it requires touching the type inference code, but if we were able to expose the declared types in a general way it would make it possible to add lots of autocomplete functionality though out the plugin.

klazuka:
Using the Diagnostics is a clever idea. It was definitely intended only for display purposes, but I guess there's no technical reason why you couldn't use it other than objections based on purity. I wouldn't be strictly opposed to it.

Changing `InferenceResult` is trickier as there can be some gotchas. The type inference data is heavily cached, and so we want to keep the size of the in-memory data small. And you have to be careful about the caching because we are caching mutable objects and promising not to mutate them after. (not sure if that would be a problem in this case, but just wanted to mention it).

You mentioned that the Diagnostics-based solution would only cover some of the situations. Which situations did you have in mind that wouldn't be covered? Where there is no type annotation?

ryancerf:
Thanks for the quick response.

Using the diagnostic solution the only place I have found so far where I would like autocomplete to work but it does not is inside a nested record. 

```elm
// no completions
type alias Foo = { name : { first: String } }

f : Foo
f =
    { name = { f{-caret-} } }
```

The `recordDiff` on the diagnostic is at the parent level (`Foo`) which makes it hard to find the missing fields on the nested record`{ first: String }`.

Here is the diagnostic from the example above.

![Screen Shot 2020-08-15 at 12 06 27 PM](https://user-images.githubusercontent.com/7135877/90319749-c57bdb00-deef-11ea-9850-18554cd84167.png)


Options
[...]

PR #682, opened 2020-08-15
Autocomplete record field names using diagnostics

ryancerf:
Addresses #680

 Does not complete field names in nested records. See #680 for more details.

This PR stops the TypeInference from short circuiting in a few situations where there are errors in the AST. I think it is necessary because incomplete records (e.g `{name }`) contain errors such as `EQ, PIPE or RIGHT_BRACE expected got }`. To be honest I do not understand all the implications of making this change.

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/completion/ElmRecordLiteralFieldSuggestor.kt
+++ b/src/main/kotlin/org/elm/lang/core/completion/ElmRecordLiteralFieldSuggestor.kt
@@ -1,0 +1,42 @@
+package org.elm.lang.core.completion
+
+import com.intellij.codeInsight.completion.CompletionParameters
+import com.intellij.codeInsight.completion.CompletionResultSet
+import com.intellij.codeInsight.lookup.LookupElementBuilder
+import org.elm.lang.core.diagnostics.TypeMismatchError
+import org.elm.lang.core.psi.ELM_IDENTIFIERS
+import org.elm.lang.core.psi.elementType
+import org.elm.lang.core.psi.elements.ElmField
+import org.elm.lang.core.psi.elements.ElmRecordExpr
+import org.elm.lang.core.types.*
+
+object ElmRecordLiteralFieldSuggestor : Suggestor {
+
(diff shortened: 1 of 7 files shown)


klazuka:
Hi @ryancerf , I was busy fixing a major bug in the plugin (#683), so I haven't had a chance to look at your PR. I will check it out in the next few days. Thanks!

src/main/kotlin/org/elm/lang/core/types/TypeInference.kt
```diff
@@ -436,6 +436,7 @@ private class InferenceScope(
             atom is ElmLetInExpr -> inferChild { beginLetInInference(atom) }.ty
             atom is ElmCaseOfExpr -> inferCase(atom)
             atom is ElmParenthesizedExpr -> inferExpression(atom.expression)
+            atom is ElmRecordExpr -> inferRecord(atom)
```
klazuka, line 439: Allowing Psi errors when inferring records worries me a little, but I haven't found any problems with it yet. By making this more relaxed, we may start to see weird type inference results caused by the Psi tree being messed up. And it may be difficult to debug.

That being said, if this feature works out, then I'd be ok with the tradeoff.

src/main/kotlin/org/elm/lang/core/completion/ElmRecordLiteralFieldSuggestor.kt
```diff
@@ -0,0 +1,42 @@
...
+                    ?.missing
+                    ?.forEach { fieldName, fieldTy ->
+                        result.add(fieldName, fieldTy, parent is ElmField)
+                    }
+        }
+    }
+}
+
+private fun CompletionResultSet.add(fieldName: String, field: Ty, inElmField: Boolean) {
+    // Do not add equals after key in a field { {-caret-} = "hello" }
```
klazuka, line 38: This comment reads strangely.

klazuka:
Also, if you want to do a video call for questions/whatever, DM me on the Elm Slack.

PR #683, opened 2020-08-17
Rewrite project/dependency loader

### LINKED ISSUE #497: ProjectLoadException when setting up elm.json
I have a running project on machine A. Then I open it on machine B, files are synced using git. elm-plugin asks me to set up my toolchain and elm.json. The former works, the latter produces the error in the screenshot.

image

For machine readability: The text is:
`Failed to attach Elm project`
`org.elm.workspace.ProjectLoadException: Could not find file /home/$USER/.elm/0.19.0/package/elm/http/2.0.0/elm.json. Is the package installed?`
(I replaced my username with $USER of course).

This error happened to me on multiple occasions, another one being loading an elm.json that is not in the uppermost directory of the project or when the sources are misplaced respectively. It pops up both on machines A and B, so I presume it is not an issue of a certain installation.
I can not figure out a workaround at this point, which is a great bummer in an otherwise great plugin.

### LINKED ISSUE #669: Unknown types for imported libraries
For some reason *some* libraries do not show the value for the final return type.

image

image

I'm happy to provide any further information you need to help debug this issue.

### LINKED ISSUE #681: Application package dependencies versions are not resolved correctly
When attempting to load the project found in [this example repo](https://github.com/vtsatskin/intellij-elm-package-bug), intellij-elm shows the following error:

Screen Shot 2020-08-14 at 4 09 42 PM

`elm make` succeeds on this project, despite the error reported by intellij-elm. I've also tried loading this project in VS Code with the elm plugin and it also worked.

Looking closely at the error message, it shows that **elm-lang**/html could not be loaded, which is odd because we would expect **elm**/html to be loaded instead, based on the project's `elm.json`. After doing some debugging on the plugin itself, I've found the issue lies in how package versions are resolved within the plugin. `kalutheo/elm-ui-explorer v8.0.0` relies on `1602/elm-feather`, which is specified by the constraint `1.0.0 <= v < 2.3.5`. So in the following code, the earliest possible version number is used (1.0.0) instead of the version being used by the application (2.3.4) as specified in the `elm.json`. 

https://github.com/klazuka/intellij-elm/blob/d37b8edce008fad409002e170cfb34d9b0295f10/src/main/kotlin/org/elm/workspace/ElmProject.kt#L260

In this specific case, we could change the `min()` to `max()` and resolve the issue. However, choosing neither the minimum or maximum version would be correct in this example because the exact version specified by the project's `elm.json` should be used as it could be any version within the constraint.

At some point in the `1602/elm-feather`'s history the package names changed which caused this issue. While the issue could be resolved by constraining the versions of `1602/elm-feather` used by `kalutheo/elm-ui-explorer`, this still doesn't address the root of the issue where the wrong version of the root project's indirect dependencies were used.

klazuka:
One of the most important things in this plugin is the logic for reading an `elm.json` file and determining the layout of source directories, the location of dependencies, and which modules are visible for a given scope. This is relatively easy to do for Elm applications because the `elm.json` file contains the concrete version numbers of each dependency, direct and indirect, whereas package projects only list _constraints_ on versions.

Early on in the project I punted on properly resolving "package project" constraints because I was hopeful that the Elm compiler would eventually be able to export information about how it resolved the constraints. That never happened, and I never got around to doing it myself.

But as it turns out, there was a deep bug in my code even for application projects which resulted in the plugin potentially using the wrong version of a package.

This PR does several things:
1. fixes the bug that caused application projects to sometimes use the wrong version of a dependency
2. implements constraint-based version resolution for package projects
3. cleans-up the code to make it easier to understand
4. gets rid of some lame hacks related to how integration tests load `elm.json` files
5. fixes go-to-declaration for indirect dependencies

Fixes #669 
Fixes #681
Fixes #497

### DIFF
--- a/src/main/kotlin/org/elm/lang/core/lookup/ElmLookup.kt
+++ b/src/main/kotlin/org/elm/lang/core/lookup/ElmLookup.kt
@@ -130,7 +130,7 @@ private fun buildSearchScope(includeTests: Boolean, p: ElmProject, intellijProje
      *        you must take that into account when retrieving the cached value.
      */
     val (srcDirPaths, dependencies) = when {
-        includeTests -> p.allSourceDirs to p.allResolvedDependencies
+        includeTests -> p.allSourceDirs to p.allDirectDeps
         else -> p.absoluteSourceDirectories.asSequence() to p.dependencies.asSequence()
     }
 
--- a/src/main/kotlin/org/elm/lang/core/lookup/ElmLookup.kt
+++ b/src/main/kotlin/org/elm/lang/core/lookup/ElmLookup.kt
@@ -11,8 +11,8 @@ import com.intellij.psi.util.CachedValueProvider
 import com.intellij.psi.util.CachedValuesManager
 import org.elm.lang.core.psi.ElmFile
 import org.elm.lang.core.psi.ElmNamedElement
-import org.elm.lang.core.stubs.index.ElmNamedElementIndex
 import org.elm.lang.core.psi.moduleName
+import org.elm.lang.core.stubs.index.ElmNamedElementIndex
 import org.elm.openapiext.findFileByPathTestAware
 import org.elm.workspace.ElmPackageProject
 import org.elm.workspace.ElmProject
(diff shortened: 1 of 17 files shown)


cmditch:
Yay! Thank you so much for all the time you put into this excellent tool.

vtsatskin:
Thanks so much @klazuka for how quickly you made this changes. I'll be using the latest build in my day-to-day programming so I'll let you know if there are any bugs from this change.

PR #691, opened 2020-08-25
Add intentions for converting between regular and triple-quoted strings.

### LINKED ISSUE #690: Add intentions for converting between regular and triple-quoted strings
It would be good if the plugin had a refactoring intention available for converting between regular strings (delimited by `"`) and triple-quoted strings (delimited by `"""`). For example to allow you to convert between:

```
example =
    "foo\nbar"
```

and

```
example =
    """foo
bar"""
```

yonigibbs:
Closes #690.

Hi Keith. As discussed here's a PR for the refactoring intentions to convert between regular and triple-quoted strings. Some points for your consideration when reviewing:
* I haven't done anything with the caret position. If we decide this should be included here, it will add some complexity, particularly in multi-line strings, so we'll need to decide if that's worthwhile or not.
* In terms of terminology, I went with "regular string" and "triple-quoted string", but happy to change this to something else (e.g. "multi-line string") if you prefer. I decided against multi-line string as it doesn't necessarily have to have multiple lines, but I'm aware that generally this is the most common usage of triple-quotes.
* I added a couple of properties into `ElmStringConstantExpr` so this class now knows a bit about delimiters. If you prefer we can move these into the intention-related code (e.g. as extension properties there). From looking at some of the other classes it looked like it was OK for this type of class to have knowledge of such things, so I went with this approach, but happy to move if you have a different preference.
* I've used `System.lineSeparator` when building up mutli-line strings: is this right, or should we just use `\n`?

@dillonkearns, fyi. Any comments/suggestions welcome as always.

### DIFF
--- a/src/main/kotlin/org/elm/ide/intentions/RegularToTripleQuotedStringIntention.kt
+++ b/src/main/kotlin/org/elm/ide/intentions/RegularToTripleQuotedStringIntention.kt
@@ -1,0 +1,38 @@
+package org.elm.ide.intentions
+
+import org.elm.lang.core.psi.elements.ElmStringConstantExpr
+import org.elm.lang.core.psi.elements.TRIPLE_QUOTE_STRING_DELIMITER
+
+/**
+ * An intention which converts regular strings to triple-quoted strings, e.g. replaces this:
+ * ```
+ *     example =
+ *         "foo\nbar"
+ * ```
+ * with this:
+ * ```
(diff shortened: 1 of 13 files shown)


klazuka:
I don't think that you need to use `System.lineSeparator`. I believe that IntelliJ normalizes everything to newlines. But in any case, your tests passed on Windows, so I think either way is fine.

PR #692, opened 2020-08-26
Auto-discover path to Elm tools and the `elm.json` file when opening a project for the first time

klazuka:
This should be a nice usability win.

- auto-discovers the Elm compiler, elm-format and elm-test (prioritizing tools installed in local `node_modules` over global install)
- auto-attaches the first `elm.json` file it finds (breadth-first search from the content root to a max depth of 3)

### DIFF
--- a/src/main/kotlin/org/elm/openapiext/Utils.kt
+++ b/src/main/kotlin/org/elm/openapiext/Utils.kt
@@ -78,16 +78,16 @@ fun fullyRefreshDirectory(directory: VirtualFile) {
     VfsUtil.markDirtyAndRefresh(/* async = */ false, /* recursive = */ true, /* reloadChildren = */ true, directory)
 }
 
-fun VirtualFile.findFileBreadthFirst(predicate: (VirtualFile) -> Boolean): VirtualFile? {
-    val queue = LinkedList<VirtualFile>().also { it.push(this) }
-    while (queue.isNotEmpty()) {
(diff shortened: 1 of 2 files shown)


pravdomil:
does not work for me?

image

klazuka:
@pravdomil what version/flavor of IntelliJ are you using?

How did you create the project?

pravdomil:
```
IntelliJ IDEA 2020.1.2
IntelliJ Elm 4.1.0
```
```sh
mkdir foo
cd foo
elm init
touch src/Main.elm
idea .
```

klazuka:
The problem is that you're opening an individual source file instead of a directory. I have the same configuration as you, but if instead of `idea src/Main.elm` you instead do `idea .` it works.

Launching IntelliJ on an individual source file appears to load it in a lightweight mode without project view, etc.

pravdomil:
oh sorry I edited the comment, I open the whole project, not a single file, see screenshot

pravdomil:
I'm on macOS 10.13.3.

klazuka:
Are you saying that on macOS 10.13.3 when you open the `foo` directory in IntelliJ (either from CLI or GUI) it doesn't automatically attach the `elm.json` file?

pravdomil:
It does not, tried many times.

klazuka:
I followed up with you on the Elm Slack. Let's take it from there.

PR #693, opened 2020-08-26
Allow elm-format to run on-save even if there are syntax errors

### LINKED ISSUE #628: Run `elm-format` when file saved even when there are syntax errors
## Problem

`elm-format` isn't executed when a file is saved, and I think this is because the plugin detects syntax errors (I am not confident in that, but that matches my experiments).

## Motivation

Coming from Atom and `elmjutsu`, I am a bit frustrated with how many more syntax errors I get now compared to with `elmjutsu`.

From what I understand, `elm-format` has looser syntax requirements than IntelliJ, but fixes them when it is run. I show a few examples where `elm-format` fixes the syntax errors below. They already get fixed today when I run `elm-format` from the CLI or from IntelliJ, so it would only be a matter of enabling it on save.

I feel that this would be a huge improvement to the user's experience, especially for newcomers to the language. With `elmjutsu` doing this this way, I never felt that the language was too strict on the syntax (like indentation for instance), but without `elm-format` covering for me, it feels more painful to write Elm.

## Examples of fixed syntax errors

(Would fix the error reported in https://github.com/klazuka/intellij-elm/issues/443 if the users enabled `elm-format` on save)
```elm
-- Doesn't compile
type alias A =
{}
-- After elm-format, compiles
type alias A =
    {}
```
(Would fix the error reported in https://github.com/klazuka/intellij-elm/issues/22 if the users enabled `elm-format` on save)
```elm
-- Doesn't compile
foo =
    a (\x ->
        case x of
            [] -> 0
            _ -> 1)
-- After elm-format, compiles
foo =
    a
        (\x ->
            case x of
                [] ->
                    0

                _ ->
                    1
        )
```
```elm
-- Doesn't compile
type alias A =
    { a = Int }
-- After elm-format, compiles
type alias A =
    { a : Int }
```
```elm
-- Doesn't compile
value =
    { a : 1 }
-- After elm-format, compiles
value =
    { a = 1 }
```
```elm
-- Doesn't compile
func a =
    let
b = 1
    in
    a
-- After elm-format, compiles
func a =
    let
        b =
            1
    in
    a
```

## Thanks

I will use this opportunity to thank you for the hard work you have put into this project. It is really a splendid tool :heart:

klazuka:
Fixes #628

### DIFF
--- a/src/main/kotlin/org/elm/ide/actions/ElmExternalFormatAction.kt
+++ b/src/main/kotlin/org/elm/ide/actions/ElmExternalFormatAction.kt
@@ -1,6 +1,5 @@
 package org.elm.ide.actions
 
-import com.intellij.execution.ExecutionException
 import com.intellij.notification.NotificationType
 import com.intellij.openapi.actionSystem.AnAction
 import com.intellij.openapi.actionSystem.AnActionEvent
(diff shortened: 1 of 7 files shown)


jfmengels:
Thank you! :heart_eyes:

PR #697, opened 2020-08-29
Complete fields in record expressions

### LINKED ISSUE #680: Autocomplete Record Field names

ajalt:
Rather than trying to infer partial programs, this PR instead uses the normal completion mechanisms to fix up the PSI so it parses correctly before finding suggestions.

A situation where I didn't add completion is like this: `{foo: a b{-caret-}}`. It's not possible to tell whether `b` is supposed to be a function call or a field name, so I left that case alone.

Fixes #680

### DIFF
--- a/src/test/kotlin/org/elm/lang/core/completion/ElmRecordExprCompletionTest.kt
+++ b/src/test/kotlin/org/elm/lang/core/completion/ElmRecordExprCompletionTest.kt
@@ -190,4 +190,14 @@ f =
     { second = { third = Baz { foo = {-caret-} } } }
 """
     )
+
+    fun `test inside tuple`() = doSingleCompletion("""
+f : ({ foo : () }, ())
+f =
+    ({ f{-caret-} }, ())
+""", """
+f : ({ foo : () }, ())
+f =
+    ({ foo = {-caret-} }, ())
+""")
 }
--- a/src/test/kotlin/org/elm/lang/core/completion/ElmRecordExprCompletionTest.kt
+++ b/src/test/kotlin/org/elm/lang/core/completion/ElmRecordExprCompletionTest.kt
@@ -80,13 +80,6 @@ f =
     { foo = {-caret-} }
 """)
 
-    fun `test leaf of nested record`() = checkNoCompletion(
-            """
-f : { foo : { bar: () } }
-f =
-    { foo = { s{-caret-} } }
-""")
-
     fun `test custom type declaration`() = doSingleCompletion(
             """
 type Bar = Baz { foo : () }
--- a/src/test/kotlin/org/elm/lang/core/completion/ElmRecordExprCompletionTest.kt
+++ b/src/test/kotlin/org/elm/lang/core/completion/ElmRecordExprCompletionTest.kt
@@ -1,0 +1,193 @@
+package org.elm.lang.core.completion
+
+class ElmRecordExprCompletionTest : ElmCompletionTestBase() {
+
+    fun `test blank`() = doSingleCompletion(
+            """
+f : { foo : () }
+f =
+    { {-caret-} }
+""", """
+f : { foo : () }
+f =
+    { foo = {-caret-} }
+""")
+
+    fun `test one letter`() = doSingleCompletion(
+            """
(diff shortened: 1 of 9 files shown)


klazuka:
1. you're awesome 💯 
2. you stole my idea 😄  I had come to the same conclusion that the best path forward would be to fix-up the syntax errors during the completion process

I'll look at this later. Thanks!

klazuka:
When running the code in the IDE and evaluating it interactively, I'm seeing the same kind of random bugs that @ryancerf encountered. For instance, in [elm-spa-example/src/Page/Blank.elm](https://github.com/rtfeldman/elm-spa-example/blob/cb32acd73c3d346d0064e7923049867d8ce67193/src/Page/Blank.elm#L8), I deleted a few characters from `content`, then did ctrl-space and it correctly suggested "content". Then I moved the cursor up one line and started deleting the `title` field. Now when I type in "ti" and try to complete the field name, it doesn't suggest anything. But it will still suggest "content".

The problem goes away if I delete the entire `ElmValueDeclaration` and type it in again. I suspect that it's related to the caching. Maybe it has something to do with the way that IntelliJ code completion creates a shadow file (where `PsiFile.originalFile()` gives you a different file than what `PsiElement.containingFile()` gives). Maybe we need to key the caches on an additional parameter so that they are isolated?

klazuka:
Another thing I noticed is that it doesn't seem to work when the record literal is in a tuple. This is important for `update` style functions which return `(Model, Msg)`.

The following test fails:
```kotlin
    fun `test inside a tuple`() = doSingleCompletion("""
f : ({ foo : (), other: () }, ())
f =
    ({ f{-caret-} }, ())
""", """
f : { foo : (), other: () }
f =
    ({ foo = {-caret-} }, ())
""")
```

src/test/kotlin/org/elm/lang/core/completion/ElmRecordExprCompletionTest.kt
```diff
@@ -0,0 +1,193 @@
...
+f : { foo : { bar: () } }
+f =
+    { f{-caret-} }
+""", """
+f : { foo : { bar: () } }
+f =
+    { foo = {-caret-} }
+""")
+
+    fun `test leaf of nested record`() = checkNoCompletion(
```
ryancerf, line 83: Is there a typo here? Should the s{-caret-} be changed to b{-caret-} to match the first letter of bar?

If this is a typo how come completions do not work inside nested records?

If this is not a typo then consider updating the name of the test.

ajalt:
I fixed the tuple problem by reporting type errors on individual tuple elements instead of the whole tuple. I think that looks better regardless of the completion.

I wasn't able to figure out the root cause of the stale completion results, so I added a hack and an apology instead.

I also improved the completion dropdown by not showing `=` in it.

src/test/kotlin/org/elm/lang/core/completion/ElmRecordExprCompletionTest.kt
ajalt: I don't even know why that test was there. 

The reason nested records don't work is the same reason that tuples didn't: we report a single error on the entire outer record rather than more granular errors. 

The fix would be the same as the fix for tuples: report errors on the individual fields. That would work for extra or mismatched fields, but not for missing. I don't know of a good way to report missing fields other than what we do now.

ryancerf: I ran into the same issue in my attempt. I started playing around with showing the more granular errors when all the expected keys in the outer record were present otherwise showing the error for the entire outer record. Do you think something like that could make sense?

ajalt: That's basically what I'm doing with tuples. I think it might feel inconsistent to do the same thing with records, but maybe it would be ok.

PR #698, opened 2020-08-29
Cleanup pipeline intentions

klazuka:
@dillonkearns I went and made some minor improvements to make the code more closely match the style of the rest of the project.

This refactoring was pretty fun to do since you had a nice suite of integration tests and you already handled the hard parts (multi-line Elm code with indentation madness).

I ran out of steam on `RightPipeline.segments`. That still needs a bit more work. I suspect that it could be made a lot simpler if you just fully go over to the `Collection` world (i.e. `List` and `MutableList`) instead of `Sequence`. Mixing and matching the different types of "collections" in Kotlin is not great. Better to just go all-in with one. And in most circumstances, that will be `Collection`/`List`, not `Sequence` (@ajalt wrote about Sequence performance [here](https://medium.com/@ajalt/benchmarking-kotlin-sequences-e06d8bb4011c))


klazuka:
Also, I tried to make each commit readable on its own, so you might want to look at them individually.

yonigibbs:
@klazuka, @dillonkearns, I was playing about a bit with this intention and I think there are some cases where it might potentially be a bit confusing to users. Take this code for example:
```
suite : Test
suite =
    describe "foo"
        [ test "bar" <|
            \_ ->
{-caret-}       1 |> Expect.equal 1
        , test "baz" <|
            \_ ->
                1 |> Expect.equal 2
        ]
```
When the caret is in position indicated above, you get two intentions: "Remove Pipes" and "Use pipeline |>". Each of these acts on a different level in the nesting of functions. "Remove Pipes" results in this:
```
suite : Test
suite =
    describe "foo"
        [ test "bar"           <-- PIPE REMOVED HERE
            (\_ ->
                1 |> Expect.equal 1
        )                      <-- (MINOR POINT: WRONG INDENTATION)
        , test "baz" <|
            \_ ->
                1 |> Expect.equal 2
        ]

```
And back at the original code if you chose instead to "Use pipeline of..." you get this:
```
suite : Test
suite =
    [ test "bar" <|            <-- (MINOR POINT: WRONG INDENTATION IN SUBSEQUENT LINES)
            \_ ->
                1 |> Expect.equal 1
        , test "baz" <|
            \_ ->
                1 |> Expect.equal 2
        ]
        |> describe "foo"      <-- PIPELINE ADDED HERE
```
So as a user, when I saw both intentions, it was a bit unclear to me what impact each one will have, where it will add pipelines, or which pipelines it will remove. Possibly also a bit confusing that both options show up at the same time? It also slightly broke that nice ability to try the intention out, then "undo" it by choosing the opposite action. I wonder... should the intention only ever act on one level of nesting, i.e. the "deepest" level? I think that might avoid the situation where both intentions are valid simultaneously.

One other minor thing is that the generated code sometimes gets the indentation a bit wrong, as you can see above. Arguably this isn't a big deal, especially if you have elm-format set to run on save.

Also just one last very tiny point: the captions are "Remove Pipes" and "Use pipeline of..." (notice the difference in casing).

klazuka:
I also found it hard to predict what the intention would do. 

In the case of remove pipeline, we could only make that available when the cursor is on '<|'. Similarly, we could suppress the "add pipeline" when remove pipeline is suggested. This would fix a lot of the ambiguity. 

@dillonkearns thoughts?

dillonkearns:
Good point about the different scopes.

My two cents is, I really like the idea of `Remove pipeline` only being offered when your cursor is on a `<|` or `|>`. With that change, I think the ambiguity wouldn't be an issue, because it's always clear what `Remove pipeline` applies to. And `Use pipeline of...` will always apply to the function call your cursor is in the context of.


For example:

```elm
decoder : Decoder CommentId
decoder =
    (CommentId
        |{-cursor-}> Decode.map
    )
        Decode.int

```

Here, you would see `Remove pipeline` because the cursor is over the `|>`. And because the entire expression is a function call expression, you would be offered the option to turn it to a pipeline like this:

```elm
decoder : Decoder CommentId
decoder =
    Decode.int
        |> (CommentId
                |> Decode.map
           )
```

It's a bit of a forced example, we could probably find a better one to consider that behavior. But hopefully that gives a more concrete picture.

As a user, I like having both options in that context because my cursor being on the pipeline doesn't make it less clear that I am also in a function call expression. Maybe that's just me, but that would be my personal preference as a user.

yonigibbs:
In the example above, I think I'd actually find the option to `Use pipeline of |>` a bit confusing, as I'd think that it already was using `|>` (i.e. `CommentId |> Decode.map`). It would take a bit of extra brainwork to figure out that the pipe I'm on is part of an expresssion which is itself part of another expression, and that it's that _outer_ expression which is going to be converted to a pipe. So I was wondering, what if the option was only ever available on the initial function whose arguments you want to pipe? This would be arguably more limiting, but maybe less ambiguous. So in the example above, you'd have to have your caret at `Decode.int` in order to get the option to `Use pipeline of |>`?

Or would that be equally confusing in the following case:

```
String.join "," (List.map String.fromInt [ 1, 2, 3 ])
```
... to ...
```
[ 1, 2, 3 ] |> List.map String.fromInt |> String.join "," 
```
... because in that case we've turned two functions (`String.join`) and `List.map` into a single chain of piped operations?

dillonkearns:
The 3 criteria I can think of for an intention are:

1) Discoverability - when you're looking for ways to rewrite some code, how intuitive is it to find it?
2) Availability - assuming you know you can do a transformation, how ergonomic is it to find the right place to invoke it from?
3) Consistency - how easy is it to predict when an intention will be available and when it won't be?
4) Ambiguity - how easy is it to tell what scope an intention will effect?

If we hide the `Use pipeline of |>` intention when the cursor is on a pipe operator, then I would be concerned about the impact that would have on points 1, 2, and 3.

For example, I was playing around with some Kotlin refactorings in an if condition. When the cursor is within the if condition, these 5 intentions are available:

Image 2020-08-30 at 10 57 44 AM

When you move the cursor to the `&&`, then a 6th becomes available (`Flip '&&'`). 

Image 2020-08-30 at 10 57 53 AM

If some of the other 5 intentions went away when I invoked the intention from the `&&` operator, I would find the tradeoff would be similar. It would have a negative impact for 1, 2, and 3. But it would make things less ambiguous (point 4).

yonigibbs:
All very good points! I had a quick try with this bit of Kotlin code: `if (a && b)` (where `a` and `b` are booleans). With the caret at `if`, things related to `if` are available. As soon as the caret is inside the brackets, those options disappear, and ones related to `&&` are made available instead. So to me that's kinda what I'm suggesting: once inside the "scope" of the `&&`, things related to `if` are no longer there. But having said that I also totally agree with everything you say above :-)

dillonkearns:
Yeah, I noticed the special cases that are available for the `if`. To me, the difference there is that the if condition is its on context. Whereas a function call expression includes the `|>` token. So presenting the `if` intentions only when on the `if` keyword can be done without suppressing it in certain cases. But to me, the idea of suppressing an intention based on context (rather than defining a clear context that doesn't overlap with other contexts) makes things less predictable.

klazuka:
Good points. I'm going to merge this PR as its purpose was solely to refactor the style. 

@yonigibbs @dillonkearns It sounds like we are agreed about one thing: change "Remove Pipeline" so that it is only available when the cursor is on a pipe operator. Maybe that change is sufficient for now and we can get feedback on the other issues once it ships?

@dillonkearns if this sounds good to you, would you mind creating a new PR with that change?

dillonkearns:
Sounds great @klazuka! I'll follow up with a PR with that change 👍

PR #700, opened 2020-08-30
Cleanup Pipeline.segments function

dillonkearns:
I'm not sure what you think of this, there may be other ways to clean it up. The main part that makes it messy is that we have to offset the comments by one.

Let me know what you think of this refactoring, I'm happy to make some more adjustments.


klazuka:
Is this functionally better than the old code? Because the new one is a lot longer, and if it's functionally equivalent, I would prefer to just keep the old code.

dillonkearns:
It's functionally the same. Let's close this and keep it for reference. If we come back to this code and are able to change the logic so it doesn't require shifting the comments around, then we can tweak this potentially to clean things up.

PR #701, opened 2020-08-31
Only show Remove Pipeline intention when the cursor is over a pipeline operator

dillonkearns:
Here is the change we discussed in #698. Let me know if you have any feedback, I'm happy to change things if you have any thoughts. Thank you!

### DIFF
--- a/src/main/kotlin/org/elm/ide/intentions/RemovePipelineIntention.kt
+++ b/src/main/kotlin/org/elm/ide/intentions/RemovePipelineIntention.kt
@@ -4,11 +4,8 @@ import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
 import com.intellij.util.DocumentUtil
-import org.elm.lang.core.psi.ElmPsiElement
-import org.elm.lang.core.psi.ElmPsiFactory
-import org.elm.lang.core.psi.ancestors
+import org.elm.lang.core.psi.*
 import org.elm.lang.core.psi.elements.*
-import org.elm.lang.core.psi.startOffset
 import org.elm.lang.core.withoutExtraParens
 import org.elm.lang.core.withoutParens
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/elm/ide/intentions/RemovePipelineIntention.kt
```diff
@@ -24,11 +21,9 @@ class RemovePipelineIntention : ElmAtCaretIntentionActionBase<RemovePipelineInte
     override fun getFamilyName() = text
 
     override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? =
-            element.ancestors
-                    .filterIsInstance<ElmBinOpExpr>()
-                    .firstOrNull()
-                    ?.asPipeline()
-                    ?.let { Context(it) }
+            element.ancestors.filter(::isPipelineOperator).firstOrNull()?.let {
```
klazuka, line 24: code formatting nits:

I prefer doing a single thing per line (unless it's really lightweight). So I would keep the formatting similar to how it was before. In general, I like Elm's spacious approach to code-formatting. I dislike dense, "wall-of-text" code.

We rarely use Kotlin's `::method` syntax. I prefer just doing `{ it.isPipelineOperator }` instead. My thinking is:
- it's one less thing that you need to know when reading the code
- if we ever need to change it to do something more complicated, the lambda will be easier to change

klazuka: Another thing about code formatting (although I don't think it's applicable to this PR):

I use Kotlin's standard code formatting settings and I press `cmd-opt-L` frequently when writing code. It's nowhere near as opinionated as elm-format, but it's at least something.

dillonkearns: Great points, agreed! How does it look now?

PR #710, opened 2020-09-05
Add more control over how types are exposed.

### LINKED ISSUE #676: Allow for exposing of custom types both with constructors and without
When exposing a custom type with `Option+Enter`, the only option is to expose the type without its constructors. While this is a good default, it would be nice to be able to expose the type with its constructors from the menu as well.

yonigibbs:
Closes #676

This PR adds the ability to choose whether to expose a type with/without variants. For a type already exposed, allows variants to be exposed/hidden.

Some points:
* I ended up with a handful of related classes, so decided to create a new package under `intentions`, called `exposure`. All intentions related to exposing/hiding stuff are now in there.
* Because of the new terminology on the UI (e.g. "Expose" instead of "Add to exposing list") and the fact that there are a few slightly similar intentions now, I renamed a couple of the existing classes (e.g. `AddExposureIntention` -> `ExposeIntention`). This was to prevent the names of the newly added intentions being overly long.
* A few of the class names are slightly similar, e.g. `ExposeIntention`, `ExposeVariantsIntention`, `ExposeWithVariantsIntention`. Let me know if you want names that are more different here. Maybe we should include the word `Type` somewhere in the latter two?
* There is some slight duplication in the tests for these, as some of the intentions are quite similar. Happy to refactor into some base test class for these, but I felt that as it was test code, for the sake of simplicity it was OK to have a small amount of duplication here.
* I found a couple of existing issues with these intentions if the current Elm code is not quite valid. For example if you have `module Main exposing ()` and use the intention, you end up with `module Main exposing (, foo)` (notice the leading comma). I didn't fix them (just checked the new code didn't make them any worse): do you want me to log these as minor issues in the repo?
* Git got slightly confused as I was renaming/moving files, and thinks I deleted `RemoveExposureIntention`. I didn't: it's been renamed to `StopExposingIntention`, and is now in the new package. Some code from it is moved up to a superclass as it was needed in a couple of places. Same with some of the templates. Sorry: the PR looks a bit bigger than it actually is.
* I've not yet done any work on the gutter icons. It's not directly related to the intentions, so I'll do that in a separate PR (or make another push into this PR, depending on when I get a chance to look at it).


yonigibbs:
I had a go at the gutter icons. First idea what to add `(..)` in a smaller font after the `e` in the current icon. That looked great in the SVG file, and absolutely terrible in the really small PNG which IntelliJ would use :-( I ended up thinking that trying to distinguish between a normal exposure and an exposure with variants might be too much for such a small number of pixels. So for the moment I left the icon as-is. Instead, I updated the tooltip text. It now says "Exposed" normally (as it used to) unless it's a custom type with its variants exposed, in which case it now says "Exposed (including variants)". Have a look and see what you think. Happy to change it any way you see fit.

klazuka (approved):
Works great! Nice job!

The class names are a little confusing, but the doc comments made it clear. It's hard to name these things, that's for sure!

As for the gutter icons, yeah, these small icons are a real pain. I'm fine with leaving it as-is.

klazuka:
As for the existing bugs, please go ahead and create issues for anything that you found.

yonigibbs:
Great stuff, thanks @klazuka. Logged a separate issue for the minor problems with invalid `exposing` syntax: #712.

Issue #661, opened 2020-04-26
Feature request: Lifting undefined functions into their own correctly typed definition


jakeyboy24:
`intToString : Int -> String`
`intToString int =`
&nbsp;&nbsp;&nbsp;&nbsp;`int |> helperFunction`

being able to option enter over `helperFunction` and yield

`helperFunction : Int -> String`
`helperFunction int =`

`intToString : Int -> String`
`intToString int =`
&nbsp;&nbsp;&nbsp;&nbsp;`int |> helperFunction`

lydell:
Possibly related: https://github.com/klazuka/intellij-elm/issues/320

PR #711, opened 2020-09-08
Create from inferred usage

dillonkearns:
For reference, here's the work-in-progress from [our live stream today](https://www.youtube.com/watch?v=GcWF-ZOLe5k). I'll update this thread as I finish these todos.

## Current State

Screen Recording 2020-09-07 at 04 55 48 PM

### Todos
- [ ] Don't suggest for unresolved type references - only for value functions (for this PR, revisit in the future)
- [ ] Don't suggest fix for qualified references
- [ ] Make sure cursor focus goes through argument names so you can tab through to change them
	* Make declaration fix does it
	* It's called a Template (see TemplateManager)
- [ ] Add newline for proper formatting for creating function
- [ ] Location - insert it right below call site function body, not at bottom

### Maybe now, maybe defer?
- [ ] Infer types based on values passed in?
- [ ] Use argument template names based on types, like create from declaration?

### DIFF
--- a/src/main/kotlin/org/elm/ide/inspections/MakeDeclarationFromUsageFix.kt
+++ b/src/main/kotlin/org/elm/ide/inspections/MakeDeclarationFromUsageFix.kt
@@ -19,15 +19,20 @@ class MakeDeclarationFromUsageFix : NamedQuickFix("Create") {
 
         val arity = (element.parent as? ElmFunctionCallExpr)?.arguments?.count() ?: 0
 
-        val hasArity0 = arity == 0
+        val argList: String = argListFromArity(arity)
         val valueDeclaration =
-       if (hasArity0) {
-           psiFactory.createTopLevelFunction("$s = Debug.todo \"TODO\"")
-       } else {
-           psiFactory.createTopLevelFunction("$s a = Debug.todo \"TODO\"")
-       }
+           psiFactory.createTopLevelFunction("$s $argList= Debug.todo \"TODO\"")
         // spliceIntoTopLevel
         element.elmFile.add(valueDeclaration)
+    }
+
+    private fun argListFromArity(arity: Int): String {
+        return when (arity) {
+            0 -> ""
+            else -> {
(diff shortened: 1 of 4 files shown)


dillonkearns:
A fun reference, Kolja shared this PR with me where he implements the same functionality for the Elm languages server:

https://github.com/elm-tooling/elm-language-server/pull/319

src/main/kotlin/org/elm/ide/inspections/MakeDeclarationFromUsageFix.kt
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        // addTopLevelFunction("greet", psiFactory)
+        val psiFactory = ElmPsiFactory(project)
+        val s = element.referenceName
+
+
+        val arity = (element.parent as? ElmFunctionCallExpr)?.arguments?.count() ?: 0
+
+        val argList: String = argListFromArity(arity)
```
ajalt, line 22: You can generate type-based parameter names like this:

```kotlin
val inference = element.findInference() ?: return
val callExpr = element.parent as? ElmFunctionCallExpr ?: return
val argsTys = callExpr.arguments.map { inference.elementType(it) }
val parameters = argsTys.joinToString(" ") { it.renderParam() }
```

And then, if `argsTys` isn't empty, you can generate a type annotation the way the [MakeAnnotationIntention does](https://github.com/klazuka/intellij-elm/blob/2d8d9902ba0bcd585fdf0af6251f34160c01c987/src/main/kotlin/org/elm/ide/intentions/MakeAnnotationIntention.kt#L43).

```kotlin
"$s : ${TyFunction(args, TyUnknown()).renderedText(elmFile = element.elmFile).replace("→", "->")}"
```

klazuka:
This will fix #661

cies:
Hi @dillonkearns ! I took over the stewardship of this repo. I find this PR nice to merge. Do you want to address the point @ajalt made and extend on this feature before I merge it?

PR #715, opened 2020-09-10
Add intention to convert record constructor function to record literal

yonigibbs:
Hi Keith. As discussed on Discord, here's the PR for this next refactoring intention. Some stuff to consider:
* The logic around the PSI structure in `RecordConstructorToLiteralIntention.findApplicableContext`: I'm possibly being overly defensive/specific here? Do you think there's a better approach, maybe using some of the extension functions in `PsiElement.kt` (e.g. all the stuff related to `ancestors`)?
* I'm not handling piped operations (e.g. `1 |> MyTypeAlias "a"`). In those situations the intention isn't available. I think that's OK, but let me know if you disagree.
* I'm not currently removing brackets which become redundant when the intention is applied. For example `MyTypeAlias "a" (1 + 2)` becomes `{ foo = "a", bar = (1 + 2) }` - the brackets around `1 + 2` are no longer required. Again, I think that's OK, but let me know if you want me to revisit that.
* I'm not checking the types passed into the constructor function are valid. The intention is available even if you're passing the wrong data types into the existing function. Obviously it'll generate code that doesn't compile (as the current code won't) but I think that's OK.

### DIFF
--- a/src/main/kotlin/org/elm/ide/intentions/RecordConstructorToLiteralIntention.kt
+++ b/src/main/kotlin/org/elm/ide/intentions/RecordConstructorToLiteralIntention.kt
@@ -1,0 +1,90 @@
+package org.elm.ide.intentions
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiWhiteSpace
+import com.intellij.psi.impl.source.tree.LeafPsiElement
+import org.elm.lang.core.psi.ElmPsiFactory
+import org.elm.lang.core.psi.ElmTypes.UPPER_CASE_IDENTIFIER
+import org.elm.lang.core.psi.elements.*
+
+/**
+ * An intention action that transforms a usage of record constructor to the corresponding record literal, i.e. converts
+ * this:
+ * ```
+ * type alias MyTypeAlias = { foo: String, bar: Int }
+ *
+ * baz = MyTypeAlias "a" 1
+ * ```
+ * to this:
+ * ```
+ * type alias MyTypeAlias = { foo: String, bar: Int }
+ *
+ * baz = { foo = "a", bar = 1 }
+ * ```
+ */
+class RecordConstructorToLiteralIntention :
+    ElmAtCaretIntentionActionBase<RecordConstructorToLiteralIntention.Context>() {
+
+    /**
+     * @param functionCall The current record constructor function call, to be replaced with a record literal.
+     * @param arguments The arguments to the record literal. Each entry in the list is a pair of strings: the first is
+     * the argument name, the second is the argument's text.
+     */
+    data class Context(val functionCall: ElmFunctionCallExpr, val arguments: List<Pair<String, String>>)
+
+    override fun getText() = "Use record literal"
+    override fun getFamilyName() = text
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
+        // This intention is valid in a situation like this:
+        //      foo = MyRecord "a" 1
+        // In order to be valid the caret must be in the text "MyRecord". If the caret is inside this word (or just
+        // before the M) then the PSI structure is as follows:
+        //      ElmFunctionCallExpr  (this is the record constructor function)
+        //          ElmValueExpr
+        //              ElmUpperCaseQID
+        //                  LeafPsiElement (the current element, elementType = UPPER_CASE_IDENTIFIER)
+        // But if the caret is just after the last "d" of "MyRecord" the structure is as follows:
+        //      ElmFunctionCallExpr  (this is the record constructor function)
+        //          ElmValueExpr
+        //          PsiWhiteSpace    (the current element, sibling of the ElmValueExpr above)
+        val functionCall = when {
+            element is LeafPsiElement &&
+                    element.elementType == UPPER_CASE_IDENTIFIER &&
+                    element.parent is ElmUpperCaseQID &&
+                    element.parent.parent is ElmValueExpr &&
+                    element.parent.parent.parent is ElmFunctionCallExpr -> element.parent.parent.parent as ElmFunctionCallExpr
+            element is PsiWhiteSpace &&
+                    element.prevSibling is ElmValueExpr &&
+                    element.parent is ElmFunctionCallExpr -> element.parent as ElmFunctionCallExpr
+            else -> return null
+        }
+
+        // If we get here, we are in a function call. Check if that function resolves to a record type alias.
+        val recordArgNames = ((functionCall.target.reference?.resolve() as? ElmTypeAliasDeclaration)
+            // The logic in the next three lines is what's in ElmTypeAliasDeclaration.isRecordAlias, but we duplicate it
+            // here to get at the actual ElmRecordType object.
+            ?.typeExpression
+            ?.allSegments
+            ?.firstOrNull() as? ElmRecordType)
+            ?.fieldTypeList
+            ?.map { field -> field.lowerCaseIdentifier.text }
+            ?: return null
+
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/elm/ide/intentions/RecordConstructorToLiteralIntention.kt
```diff
@@ -0,0 +1,90 @@
...
+        // before the M) then the PSI structure is as follows:
+        //      ElmFunctionCallExpr  (this is the record constructor function)
+        //          ElmValueExpr
+        //              ElmUpperCaseQID
+        //                  LeafPsiElement (the current element, elementType = UPPER_CASE_IDENTIFIER)
+        // But if the caret is just after the last "d" of "MyRecord" the structure is as follows:
+        //      ElmFunctionCallExpr  (this is the record constructor function)
+        //          ElmValueExpr
+        //          PsiWhiteSpace    (the current element, sibling of the ElmValueExpr above)
+        val functionCall = when {
```
klazuka, line 53: IntelliJ does have an abstraction for matching "patterns" against the Psi tree to detect certain conditions (see `com.intellij.patterns.PlatformPatterns`), but I found that API confusing. I suspect that it suffers from its Java heritage, and that had it been written for Kotlin, it would work differently. That being said, I've never used it, so if you find it useful, I'd be interested in learning more.

In this case, I would probably just search upwards using the `ancestors` sequence helper functions that you found in our project and search for an `ElmFunctionCallExpr`. Then I would return early if `funcCallExpr.target != element` (or something like that--you'll have to fiddle with it).

src/main/kotlin/org/elm/ide/intentions/RecordConstructorToLiteralIntention.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    override fun getText() = "Use record literal"
+    override fun getFamilyName() = text
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
+        // This intention is valid in a situation like this:
+        //      foo = MyRecord "a" 1
+        // In order to be valid the caret must be in the text "MyRecord". If the caret is inside this word (or just
+        // before the M) then the PSI structure is as follows:
+        //      ElmFunctionCallExpr  (this is the record constructor function)
```
klazuka, line 45: I like this comment. This is the sort of thing that really helps months/years from now.

src/test/kotlin/org/elm/ide/intentions/RecordConstructorToLiteralIntentionTest.kt
```diff
@@ -0,0 +1,213 @@
+package org.elm.ide.intentions
+
+class RecordConstructorToLiteralIntentionTest : ElmIntentionTestBase(RecordConstructorToLiteralIntention()) {
```
klazuka, line 3: Can you add a test for a multi-line record constructor call?

e.g.
```elm
type alias MyTypeAlias = { a : String, b : Int }
foo =
    MyTypeAlias
        "A"
        1
```

Unfortunately we don't have a good way of reformatting these multi-line expressions, so I'd be fine with just making the intention unavailable when the func call expr contains a newline. Or you can try to make it work. I started to work on an abstraction to make this easier (`IndentedTextBuilder`), but it is only useful in some cases. You can see an example in `MapToFoldIntention`.

klazuka (changes requested):
Sorry for the delay in reviewing this. Overall looks great.

I left a comment about multi-line expressions. That can be a real pain, so, seriously, feel free to just disable the intention in these cases if it ends up being hard.

src/main/kotlin/org/elm/ide/intentions/RecordConstructorToLiteralIntention.kt
yonigibbs: I had a look at `PlatformPatterns` and I agree with you: it's not the most intuitive. I looked at just using `ancestors` the way you suggested but the problem was that the intention was then available if the caret is anywhere in the "scope" of that function (i.e. when it's inside any of the function's parameters). For example: `val test = MyTypeAlias "a" {-caret-}1`. Personally I find this a bit confusing: I think it's a bit clearer if the intention is only ever available if you're on the actual constructor function name itself (`MyTypeAlias` in this example). So here's what I came up with:
```suggestion
val functionCall =
    // First condition above...
    (element.ancestors.firstOrNull { it is ElmValueExpr && it.parent is ElmFunctionCallExpr }?.parent
    // Second condition above...
        ?: element.parent.takeIf { it is ElmFunctionCallExpr && element.prevSibling is ElmValueExpr }
        ?: return null)
            as ElmFunctionCallExpr
```
I'm actually not sure that's any clearer than the original code (the indentation IntelliJ applies also doesn't seem great), though the result seems to be the same. So my gut feeling on this is to keep the code as I originally had it. What d'you think?

src/test/kotlin/org/elm/ide/intentions/RecordConstructorToLiteralIntentionTest.kt
yonigibbs: Been having a look at this... the current code will still allow the intention for your example above, but will generate a single-line record literal, i.e. `foo = { a = "A", b = 1 }`. This is OK, but it would probably be better as a multi-line record literal. I got that working, and my tests all pass, with two exceptions: if your code isn't quite correctly formatted, the `IndentedTextBuilder` can end up generating invalid PSI structures. For example this works:
```
foo =
    MyType{-caret-}Alias
        "A"
        1
```
This becomes:
```
foo =
    { a = "A"
    , b = 1
    }
```
However the following will result in an invalid PSI structure:
```
foo = MyType{-caret-}Alias
    "A"
    1
```
I'll investigate more when I get a chance, but if it looks too hairy, would you rather just make it generate a single-line record literal, or would you prefer the whole intention to be disabled on multi-line functions? (I think I'd tend towards allowing the intention, but making it generate a single-line record literal.)

klazuka (approved):
Let's merge it as-is. In terms of multi-line expressions, my example was poor. I should have made the example include an **argument** which spans multiple lines. For instance, if you were to call a record constructor with a `List` field and you pass it a `List` literal that spans multiple lines. This is probably pretty rare, and it can always be dealt with after this PR.

src/main/kotlin/org/elm/ide/intentions/RecordConstructorToLiteralIntention.kt
klazuka: I agree that the alternative code was not much clearer.

As for the intention only being available when the cursor is on the record constructor itself, yes, totally! What I meant was that you can do it in 2 steps:

1. Search up through the ancestors for a function call
2. Use the `target` property on the `ElmFuncCallExpr` that you found in step 1 to verify that it (the "target") is the element that the intention was invoked on.

yonigibbs:
Cool, sounds good. Thanks for merging. Couple of final points:
1. When using `ancestors` to find an `ElmFunctionCallExpr` if you then get its `target`, that actually points at the `ElmValueExpr` (which isn't the current `element`). So the test we'd have to use is `function.target == element.parent.parent` (if the caret is inside the function name) or `function.target == element.parent` (if the caret is in the whitespace at the end of the function name). So I wondered... could we consider creating a simple API (using reified generics) to check for a specific structure? Something like `element.prevSiblingOfType<ElmValueExpr>()?.parentOfType<ElmFunctionCallExpr>()?...`. We could also make those functions take in an optional lambda where you could add some extra tests (e.g. checking for a specific `elementType` as I did in one bit), though arguably `takeIf` gives you the same power. It might make the code a bit more friendly. If I look into that would that be something you might consider merging?
2. I'll look more into the multi-line stuff when I get a chance. I agree that it would definitely be a more friendly refactor if we could maintain line breaks.

PR #716, opened 2020-09-13
Only show New Elm File option inside Elm source directories

ajalt:
It's a bit annoying to have the "New Elm File" action show up even in projects that don't use Elm at all. This PR changes that action so that it only appears inside Elm source directories.

[Kotlin](https://github.com/intellij-rust/intellij-rust/blob/3f4eb6e80c08ee8c987e321ea5abedab8f71e4f7/src/main/kotlin/org/rust/ide/actions/RsCreateFileAction.kt#L25) and [Rust](https://github.com/JetBrains/kotlin/blob/a7f57646bf55aff3dcfd8f2c313e1ced36c2e9a5/idea/src/org/jetbrains/kotlin/idea/actions/NewKotlinFileAction.kt#L126) do something similar.

### DIFF
--- a/src/main/kotlin/org/elm/ide/actions/ElmCreateFileAction.kt
+++ b/src/main/kotlin/org/elm/ide/actions/ElmCreateFileAction.kt
@@ -2,8 +2,12 @@ package org.elm.ide.actions
 
 import com.intellij.ide.actions.CreateFileFromTemplateAction
 import com.intellij.ide.actions.CreateFileFromTemplateDialog
+import com.intellij.openapi.actionSystem.CommonDataKeys
+import com.intellij.openapi.actionSystem.DataContext
+import com.intellij.openapi.actionSystem.PlatformDataKeys
 import com.intellij.openapi.diagnostic.logger
 import com.intellij.openapi.project.Project
+import com.intellij.openapi.vfs.VfsUtil
 import com.intellij.psi.PsiDirectory
 import com.intellij.psi.PsiFile
 import org.elm.lang.core.ElmFileType
@@ -57,6 +61,17 @@ class ElmCreateFileAction : CreateFileFromTemplateAction(CAPTION, "", ElmFileTyp
     fun testHelperCreateFile(name: String, dir: PsiDirectory): PsiFile? =
             createFile(name, ELM_MODULE_KIND, dir)
 
+    override fun isAvailable(dataContext: DataContext): Boolean {
+        if (!super.isAvailable(dataContext)) return false
+
+        val file = CommonDataKeys.VIRTUAL_FILE.getData(dataContext) ?: return false
+        val project = PlatformDataKeys.PROJECT.getData(dataContext)!!
+        val sourceDirs = project.elmWorkspace.allProjects.asSequence()
+                .flatMap { it.allSourceDirVirtualFiles.asSequence() }
+        return sourceDirs.any {
+            VfsUtil.isAncestor(it, file, /*strict=*/ false)
+        }
+    }
 
     private companion object {
         private val CAPTION = "New Elm file"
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/elm/ide/actions/ElmCreateFileAction.kt
```diff
@@ -57,6 +61,17 @@ class ElmCreateFileAction : CreateFileFromTemplateAction(CAPTION, "", ElmFileTyp
     fun testHelperCreateFile(name: String, dir: PsiDirectory): PsiFile? =
             createFile(name, ELM_MODULE_KIND, dir)
 
+    override fun isAvailable(dataContext: DataContext): Boolean {
+        if (!super.isAvailable(dataContext)) return false
+
+        val file = CommonDataKeys.VIRTUAL_FILE.getData(dataContext) ?: return false
+        val project = PlatformDataKeys.PROJECT.getData(dataContext)!!
+        val sourceDirs = project.elmWorkspace.allProjects.asSequence()
```
ajalt, line 69: I'm not very familiar with the ElmWorkspace code, so let me know if there's a better way to get his info.

src/main/kotlin/org/elm/workspace/ElmProject.kt
```diff
@@ -84,6 +87,20 @@ sealed class ElmProject(
...
+    private val sourceDirCache = AtomicReference<List<VirtualFile>>()
+
+    /**
+     * Returns [VirtualFile]s for each directory in [allSourceDirs].
+     */
+    val allSourceDirVirtualFiles: List<VirtualFile>
+        get() {
+            val cached = sourceDirCache.get()
+            if (cached?.all { it.isValid } == true) return cached
+            val dirs = allSourceDirs.mapNotNull { LocalFileSystem.getInstance().findFileByIoFile(it.toFile()) }.toList()
```
ajalt, line 99: This takes a couple of milliseconds to run, and `ElmCreateFileAction.isAvailable` is called frequently (e.g. every time you click in the project pane), so it's worth caching.

src/main/kotlin/org/elm/ide/actions/ElmCreateFileAction.kt
klazuka: Use `findProjectForFile()`

this will do it:
```kotlin
return project.elmWorkspace.findProjectForFile(file) != null
```

src/main/kotlin/org/elm/workspace/ElmProject.kt
klazuka: This might no longer be necessary if you make the change suggested above. `ElmWorkspaceService.findProjectForFile()` uses an index (forked from intellij-rust) which should be pretty efficient. But you should double-check the actual performance difference.

klazuka (changes requested):
This is a nice change. I'd love to do the same for the "tool windows" (the thin button stripe around the perimeter of the IntelliJ window), but last I checked, there was no API to make it contextual.

src/main/kotlin/org/elm/ide/actions/ElmCreateFileAction.kt
ajalt: Yep, that's perfect. The performance is a couple of microseconds.

src/main/kotlin/org/elm/ide/actions/ElmCreateFileAction.kt
```diff
@@ -57,10 +60,17 @@ class ElmCreateFileAction : CreateFileFromTemplateAction(CAPTION, "", ElmFileTyp
...
+
+        val file = CommonDataKeys.VIRTUAL_FILE.getData(dataContext) ?: return false
+        val project = PlatformDataKeys.PROJECT.getData(dataContext)!!
+        return project.elmWorkspace.findProjectForFile(file) != null
+    }
 
     private companion object {
-        private val CAPTION = "New Elm file"
-        private val ELM_MODULE_KIND = "Elm Module" // must match name of internal template stored in JAR resources
+        private const val CAPTION = "New Elm file"
```
ajalt, line 72: Just fixing warnings here.

klazuka (approved):
Thanks!

PR #728, opened 2020-12-07
Adds an intention for inserting Debug.log statements

KasMA1990:
In my local setup, I've introduced this live template:

    (Debug.log "$CONT$" ($CONT$))

I'll cut out the code I want to debug, type "dbg" to insert this template, and paste to get the statement I cut out. This has been immensely useful to me, and I figured I could do one better if this was an intention in the plugin. I hope you'll find it a worthy piece of functionality to include :slightly_smiling_face: 

In it's current form, you can add log statements for any value (as far as I can tell), with a few special cases:

 - Executing the intent on a value that is part of a binary operation will select the entire operation for debugging.
      - Unless the binary operation is function composition, in which case the operation is not a valid target.
 - Executing the intent on a function name will check if the function is being called, and then the result of the call to that function is what will be logged.
 - Executing the intent on any part of a `case` statement that is not the result of a branch or the input being matched, will wrap the entire `case` in a debug.
 - The intent does not support adding log statements inside a pipeline/composition at the moment. This would be very nice to support, but I've left that as future work.

I'm unsure about the handling of the function composition. There's nothing wrong with supporting it, but it simply doesn't give a sensible output. You can still use the intent on e.g. callbacks being passed around in the code, so it's a bit inconsistent. But if a user wants to debug something in a composition, I'm guessing what they really want, is to insert a `Debug.log` inside the composition (which is not supported at the moment). So ignoring compositions is just this weird stop-gap.

This is the first time I've contributed to an IntelliJ plugin and also the first time I've written Kotlin code, so feel free to pick any nits that you feel are worth picking; there's likely to be a bunch of things I've missed. I also don't know if my test coverage is good enough, so feel free to point out good cases that would be worth looking into.

As more future work, I would also like to do an "undebug" intention at some point, but that's for another time :slightly_smiling_face:

### DIFF
--- a/src/main/kotlin/org/elm/ide/intentions/InlineDebugIntention.kt
+++ b/src/main/kotlin/org/elm/ide/intentions/InlineDebugIntention.kt
@@ -1,0 +1,95 @@
+package org.elm.ide.intentions
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import org.elm.lang.core.psi.*
+import org.elm.lang.core.psi.elements.*
+import org.elm.utils.getIndent
+
+/**
+ * This intention is built to easily allow users to add log statements on values.
+ * It is currently restricted from working fully with pipelines and function composition.
+ * For pipelined operations, you can only log the complete statement (so logging the final output, in other words),
+ * and function composition is ignored entirely
+ * (since a composition produces a function, which can't be sensibly logged).
+ *
+ * Ideally, we would support inserting Debug.log statements inside a pipeline/composition,
+ * to allow users to print the data getting piped along.
+ */
+class InlineDebugIntention : ElmAtCaretIntentionActionBase<InlineDebugIntention.Context>() {
+    data class Context(val valueToDebug: ElmPsiElement)
+
+    override fun getText() = "Log this value to console"
+    override fun getFamilyName() = text
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
+        element.ancestors.forEach {
+            when (it) {
+                is ElmValueExpr -> {
+                    val functionCallParent = it.parentOfType<ElmFunctionCallExpr>()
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/elm/ide/intentions/InlineDebugIntention.kt
```diff
@@ -0,0 +1,95 @@
...
+        val debuggedText =
+                if (textLines.size > 1) {
+                    val baseIndent = editor.getIndent(valueExpr.startOffset)
+                    val indentation = "    "
+                    val indentedLines = textLines.mapIndexed { index, line -> baseIndent + indentation.repeat(index) + line }
+                    """Debug.log
+$baseIndent    "${textLines.first()} ..."
+$baseIndent    (
+$baseIndent    ${indentedLines.joinToString("\n")}
+$baseIndent    )"""
```
KasMA1990, line 88: I'm unsure about how to best specify the indentation here

KasMA1990:
For future work, I'm thinking the ability to insert a `Debug.log` statement into a pipeline/composition should be in one separate intention. Having the insert intention be separate from this one, will allow users to choose which they want when dealing with a piped statement.

Making "undebugging" a separate intention also seems reasonable. I should probably update this PR, such that you are not shown this intention ("log this thing") when you see available intentions on a value that is already being logged.

klazuka:
I just started a new job. I will try to look at this on the weekend (I'm putting a post-it note on my desk so I don't forget!)

KasMA1990:
Thanks for the heads-up, and no worries 😊

src/main/kotlin/org/elm/ide/intentions/InlineDebugIntention.kt
klazuka: you and me both. It's kinda terrible. Doing any kind of code-gen that involves multiple lines is a pain.

klazuka (approved):
Thanks!

The intention needs to be registered in `plugin.xml`, but I will do that after merging.

PR #731, opened 2020-12-28
Restricts the InlineDebugIntention from showing in several cases

KasMA1990:
I recommend reviewing this commit by commit.

To provide a better user experience, the intention should not be shown as an option to the user, if it cannot do something valuable. This commit restricts the intention from showing in these cases:

 * On string, numeric, and char literals (unless the literal is part of a binary operator expression, in which case, THAT expression is logged).
 * On calls and arguments to Debug functions

This PR also adds dynamic text to the intention, to better tell the user what it is that will be debugged when they apply the intention. This is also a prelude of sorts for removing `Debug.log` statements. This intention can now detect if a statement is already being `Debug.log`ed, so in the future, I can let it remove the `Debug.log` in that case, instead of just making the intention unavailable.

### DIFF
--- a/src/main/kotlin/org/elm/ide/intentions/InlineDebugIntention.kt
+++ b/src/main/kotlin/org/elm/ide/intentions/InlineDebugIntention.kt
@@ -20,47 +20,92 @@ import org.elm.utils.getIndent
 class InlineDebugIntention : ElmAtCaretIntentionActionBase<InlineDebugIntention.Context>() {
     data class Context(val valueToDebug: ElmPsiElement)
 
-    override fun getText() = "Log this value to console"
     override fun getFamilyName() = text
 
     override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
-        element.ancestors.forEach {
-            when (it) {
+        element.ancestors.forEach { currentExpr ->
+            when (currentExpr) {
+                is ElmConstantTag -> {
+                    // A constant is unlikely to be interesting to debug.
+                    // It is more likely the context the constant is used in, that is interesting.
+                    currentExpr.parentOfType<ElmBinOpExpr>()
+                            ?.let { parentBinOpExpr -> return Context(parentBinOpExpr) }
+                    return null
+                }
+                is ElmParenthesizedExpr -> {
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/elm/ide/intentions/InlineDebugIntention.kt
```diff
@@ -24,45 +24,91 @@ class InlineDebugIntention : ElmAtCaretIntentionActionBase<InlineDebugIntention.
...
 
     override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
-        element.ancestors.forEach {
-            when (it) {
+        element.ancestors.forEach { currentExpr ->
+            when (currentExpr) {
+                is ElmConstantTag -> {
+                    // A constant is unlikely to be interesting to debug.
+                    // It is more likely the context the constant is used in, that is interesting.
+                    currentExpr.parentOfType<ElmBinOpExpr>()
```
KasMA1990, line 32: I would like to establish some predictable rules for when the intention does what. I'm not sure the user finds it intuitive that using the intention on a literal will include a binary operation if one is present instead. Should this "promotion" also happen if the literal is used in a function call for example? Or should it just not happen?

KasMA1990:
@klazuka have you had a chance to look at this? 🙂

cies:
Hi @KasMA1990 ! I took over this project. Do you still want to get this PR merged?

KasMA1990:
Hey, yes, I think that would be nice, but it still needs reviewing 😊

cies:
I'm fairly new to the codebase, but it LGTM. :)

cies:
@KasMA1990 Could you merge `master` into that branch? I fixed the CI :)

KasMA1990:
Done, though I can't really tell if the CI is happy or not 😅

cies:
@KasMA1990 There are lot's of flaky tests at this point ( #821 ), that messes with CI.

I've just merged your branch in my `5.0.1` branch and the tests looked fine (some flakiness accepted).

Just want to let you know we have a Slack for the plugin, maybe see you there!

https://elmlang.slack.com/app_redirect?channel=intellij-plugin

PR #732, opened 2020-12-29
Improves handling of parentheses in IncompletePatternInspection

KasMA1990:
When using the inspection to add missing branches to my `case` statements, I've often run into annoyances when my `case` is wrapped in parentheses in some way. This PR improves handling of parentheses in some cases, but it's also incomplete, and I need some help to get it where I would like it to go. I've added test cases and described how each of them have changed from the old code to the new. The commented-out tests are ones that fail, and I would like to fix.

The primary reason I couldn't get some tests to work, is that there seems to be an issue with the parser. All the failing cases involve an incomplete `case` statement being wrapped in parentheses, with the closing parenthesis being right next to the final character in the incomplete `case`. The basic test is that of `(case it of)`, where the parser produces an `ElmParenthesizedExpr` wrapping an `ElmCaseOfExpr` as expected, but the closing parenthesis is included in the `ElmCaseOfExpr`, like so:

    caseOfExpr.text == "case it of)"

It's the same issue with the other failing tests, where the closing parenthesis sits at the end of a branch. Note that having the closing parenthesis at the end of the branch is actually valid Elm code, so the parser is also rejecting valid Elm code as it is now. I tried diving into it, but that was also a rabbit hole too far for me at the moment. If you could provide some guidance on what to do, I could maybe give it a shot :slightly_smiling_face: I think it's also fair to reject this work, if you don't think it's worth the cost of making that change.

PS: Sorry about laying all these PRs on you during the holidays. Don't worry about timing, I'll wait for whenever it suits you :blush:

### DIFF
--- a/src/main/kotlin/org/elm/ide/inspections/MissingCaseBranchAdder.kt
+++ b/src/main/kotlin/org/elm/ide/inspections/MissingCaseBranchAdder.kt
@@ -1,13 +1,9 @@
 package org.elm.ide.inspections
 
-import com.intellij.application.options.CodeStyle
 import com.intellij.psi.PsiDocumentManager
 import org.elm.lang.core.lookup.ElmLookup
 import org.elm.lang.core.psi.ElmPsiFactory
-import org.elm.lang.core.psi.elements.ElmAnythingPattern
-import org.elm.lang.core.psi.elements.ElmCaseOfExpr
-import org.elm.lang.core.psi.elements.ElmTypeDeclaration
-import org.elm.lang.core.psi.elements.ElmUnionPattern
+import org.elm.lang.core.psi.elements.*
 import org.elm.lang.core.psi.indentStyle
 import org.elm.lang.core.psi.oneLevelOfIndentation
 import org.elm.lang.core.psi.startOffset
@@ -65,16 +61,25 @@ class MissingCaseBranchAdder(val element: ElmCaseOfExpr) {
         var start = elements.first().prevSibling.prevSibling
 
         // Add the virtual open token if there isn't one already
-        if (existingBranches.isNotEmpty()) {
+        if (existingBranches.isNotEmpty() && !element.text.endsWith('\n')) {
             start = start.prevSibling
         }
 
         element.addRange(start, elements.last())
 
+        val parent = element.parent
+
         // Without a value in the last branch, the virtual end token ends up before the trailing
         // whitespace, so we can't grab it from the factory element and need to add it separately.
-        val trailingWs = factory.createElements("\n$existingIndent$indent$indent")
-        element.addRange(trailingWs.first(), trailingWs.last())
+        if (parent is ElmParenthesizedExpr) {
+            val trailingWs = factory.createElements("\n$indent$indent")
+            element.addRange(trailingWs.first(), trailingWs.last())
+
+            parent.replace(factory.createParens(element.text, existingIndent))
+        } else {
+            val trailingWs = factory.createElements("\n$existingIndent$indent$indent")
+            element.addRange(trailingWs.first(), trailingWs.last())
+        }
     }
 
     private fun calcMissingBranches(): Result {
--- a/src/test/kotlin/org/elm/ide/inspections/ElmIncompletePatternInspectionTest.kt
+++ b/src/test/kotlin/org/elm/ide/inspections/ElmIncompletePatternInspectionTest.kt
@@ -90,6 +90,137 @@ foo it =
             --EOL
 """)
 
+//    fun `test no existing branch in parenthesis`() = checkFixByText("Add missing case branches", """
+//type Foo = Bar | Baz | Qux
+//
+//foo : Foo -> ()
+//foo it =
+//    (<error>case{-caret-}</error> it of<error>)</error>
+//""", """
+//type Foo = Bar | Baz | Qux
+//
+//foo : Foo -> ()
+//foo it =
+//    (case it of
+//        Bar ->
+//            --EOL
+//
+//        Baz ->
+//            --EOL
+//
+//        Qux ->
+//            --EOL
+//    )
+//""")
+
+//    fun `test one existing branch in parenthesis`() = checkFixByText("Add missing case branches", """
+//type Foo = Bar | Baz | Qux
+//
+//foo : Foo -> ()
+//foo it =
+//    (<error>case{-caret-}</error> it of
+//        Baz ->
+//            ()<error>)</error>
+//""", """
+//type Foo = Bar | Baz | Qux
+//
+//foo : Foo -> ()
+//foo it =
+//    (case it of
+//        Baz ->
+//            ()
+//
+//        Bar ->
+//            --EOL
+//
+//        Qux ->
+//            --EOL
+//    )
+//""")
+
+    fun `test one existing branch in parenthesis on new line`() = checkFixByText("Add missing case branches", """
+type Foo = Bar | Baz | Qux
+
+foo : Foo -> ()
+foo it =
+    (<error>case{-caret-}</error> it of
+        Baz ->
+            ()
+    )
+""", """
+type Foo = Bar | Baz | Qux
+
+foo : Foo -> ()
+foo it =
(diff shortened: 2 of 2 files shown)


src/main/kotlin/org/elm/ide/inspections/MissingCaseBranchAdder.kt
```diff
@@ -65,16 +61,25 @@ class MissingCaseBranchAdder(val element: ElmCaseOfExpr) {
         var start = elements.first().prevSibling.prevSibling
 
         // Add the virtual open token if there isn't one already
-        if (existingBranches.isNotEmpty()) {
+        if (existingBranches.isNotEmpty() && element.text[element.textLength - 1] != '\n') {
```
KasMA1990, line 64: Only add a new line if it's actually needed

src/main/kotlin/org/elm/ide/inspections/MissingCaseBranchAdder.kt
```diff
@@ -65,16 +61,25 @@ class MissingCaseBranchAdder(val element: ElmCaseOfExpr) {
...
         element.addRange(start, elements.last())
 
+        val parent = element.parent
+
         // Without a value in the last branch, the virtual end token ends up before the trailing
         // whitespace, so we can't grab it from the factory element and need to add it separately.
-        val trailingWs = factory.createElements("\n$existingIndent$indent$indent")
-        element.addRange(trailingWs.first(), trailingWs.last())
+        if (parent is ElmParenthesizedExpr) {
+            val trailingWs = factory.createElements("\n$indent$indent")
```
KasMA1990, line 75: Indentation is slightly different in this case over the other.

src/test/kotlin/org/elm/ide/inspections/ElmIncompletePatternInspectionTest.kt
```diff
@@ -90,6 +90,137 @@ foo it =
...
+//
+//        Bar ->
+//            --EOL
+//
+//        Qux ->
+//            --EOL
+//    )
+//""")
+
+    fun `test one existing branch in parenthesis on new line`() = checkFixByText("Add missing case branches", """
```
KasMA1990, line 141: On `master`, this test fails with this diff:
image

src/test/kotlin/org/elm/ide/inspections/ElmIncompletePatternInspectionTest.kt
```diff
@@ -90,6 +90,137 @@ foo it =
...
+
+        Bar ->
+            --EOL
+
+        Qux ->
+            --EOL
+    )
+""")
+
+//    fun `test two existing branches in parenthesis`() = checkFixByText("Add missing case branches", """
```
KasMA1990, line 167: On `master`, this test fails with this diff:
image

src/test/kotlin/org/elm/ide/inspections/ElmIncompletePatternInspectionTest.kt
```diff
@@ -90,6 +90,137 @@ foo it =
             --EOL
 """)
 
+//    fun `test no existing branch in parenthesis`() = checkFixByText("Add missing case branches", """
```
KasMA1990, line 93: On `master`, this test fails with this diff:
image

src/test/kotlin/org/elm/ide/inspections/ElmIncompletePatternInspectionTest.kt
```diff
@@ -90,6 +90,137 @@ foo it =
...
+//type Foo = Bar | Baz | Qux
+//
+//foo : Foo -> ()
+//foo it =
+//    (<error>case{-caret-}</error> it of<error>)</error>
```
KasMA1990, line 98: The last `error` here was required for the test text to be parsed. I assume it's part of the same parsing problem I described in the PR text.

src/test/kotlin/org/elm/ide/inspections/ElmIncompletePatternInspectionTest.kt
```diff
@@ -90,6 +90,137 @@ foo it =
...
+//
+//        Baz ->
+//            --EOL
+//
+//        Qux ->
+//            --EOL
+//    )
+//""")
+
+//    fun `test one existing branch in parenthesis`() = checkFixByText("Add missing case branches", """
```
KasMA1990, line 116: On `master`, this test fails with this diff:
image

src/test/kotlin/org/elm/ide/inspections/ElmIncompletePatternInspectionTest.kt
```diff
@@ -90,6 +90,137 @@ foo it =
...
+//
+//        Qux ->
+//            ()
+//
+//        Bar ->
+//            --EOL
+//    )
+//""")
+
+    fun `test two existing branches in parenthesis on new line`() = checkFixByText("Add missing case branches", """
```
KasMA1990, line 195: On `master`, this test fails with this diff:
image

src/test/kotlin/org/elm/ide/inspections/ElmIncompletePatternInspectionTest.kt
KasMA1990: On my branch, this test fails with this diff:
image

KasMA1990: Arguably, the new code produces worse output here. Using this inspection on a text ala `(case it of)` may also seem somewhat contrived, but I often find that I need a `case` inside parentheses, so it's very natural for me to start with the parentheses, and fill in the code from there. For me, this case is really the one I would like to fix.

src/test/kotlin/org/elm/ide/inspections/ElmIncompletePatternInspectionTest.kt
KasMA1990: On my branch, this test fails with this diff:
image

src/test/kotlin/org/elm/ide/inspections/ElmIncompletePatternInspectionTest.kt
KasMA1990: On my branch, this test fails with this diff:
image

klazuka:
Sorry, I just started a new job and haven't been paying attention to open-source PRs.

There has been a long-standing bug in the parser related to `case` expressions wrapped in parens. I never looked into it, but if I had to guess, I would say that the problem is related to how whitespace-sensitive parsing is done. We use GrammarKit to parse the source text. But it is designed for languages where whitespace is ignored. To workaround this, we have a [special lexer](https://github.com/klazuka/intellij-elm/blob/master/src/main/kotlin/org/elm/lang/core/lexer/ElmLayoutLexer.kt) which detects significant whitespace changes and emits virtual tokens. The parser then uses these virtual tokens to help delimit things like `case` expressions, `let` and function bodies.

src/main/kotlin/org/elm/ide/inspections/MissingCaseBranchAdder.kt
AleksandrSl, line 64:
```suggestion
        if (existingBranches.isNotEmpty() && element.text.endsWith('\n')) {
```
Would this work the same way?

or 
```
        if (existingBranches.isNotEmpty() && element.text.last() != '\n') {
```

KasMA1990: Ah, thanks for pointing that out. I'm new to Kotlin, so I missed those parts of the API :slightly_smiling_face:

KasMA1990:
No worries @klazuka. I'm happy about getting a natural break from the work here too, so it's really no issue. I'll try and poke around in the lexer and see what comes up :blush:

cies:
Can I close this issue?

KasMA1990:
Yeah, that's fine, I don't plan to continue work on it 🙂

PR #737, opened 2021-02-14
[WIP] Method extraction

ishmum123:


ishmum123:
retest this please

klazuka (changes requested):
Thanks. This has the potential to be an awesome feature, and I like where you're going with it. 

Please add some additional tests for more complex situations. I tried it out on [an expression in elm-spa-example ](https://github.com/rtfeldman/elm-spa-example/blob/cb32acd73c3d346d0064e7923049867d8ce67193/src/Page/Profile.elm#L73-L76), and it generated bad code (it didn't extract a function parameter for `maybeCred`).

klazuka:
I tried out your latest branch on elm-spa-example `src/Page/Profile.elm`. Attached is a video showing (1) how I perform ad-hoc testing of features like this to verify that it generates correct code and (2) shows several places where the method extraction does not work correctly (although I should say that in many cases it works well).

The main issues I found are:
1. bad code gen when the expression is nested in `case` expressions
2. one case where a value from the lexical scope did not have a function parameter (see video at 1:40)

These issues would need to be fixed before I would merge it. My basic criteria for any refactoring is that if the refactoring is offered by the plugin, then invoking the refactoring should always generate correct code.

Another small issue is that generated functions always have the name `fn`, even if one already exists. It'd be nice if it added a numeric suffix as needed.

https://user-images.githubusercontent.com/84525/113512095-1567ea80-9531-11eb-8031-edbd4bcfe359.mov

ishmum123:
Thank you so much for the response. Would try to get it done ASAP.

ishmum123:
@klazuka I think the current push fixes the issues you highlighted. Sorry for making you check so much... I would need to give some thought for making the enumerated `fn`. Please let me know if that is absolutely necessary

ishmum123:
Here's a check with the exact examples you used... Can you please suggest something about the `feed` case?

https://user-images.githubusercontent.com/9882721/116026676-fdd5da80-a674-11eb-860b-47111db65917.mov

Issue #746, opened 2021-04-09
fnm-managed node requires Elm Toolchain auto discover on every start


adamnfish:
I'm using `fnm` to manage node versions and libraries on my development machine. Following the `npm` instructions works fine, with the following caveat.

`fnm` does not use consistent paths for the environment so every time IntelliJ is started after a reboot/login it needs to be re-shown the paths to the elm compiler (and elm format/test). Hitting `Auto Discover` for each of them works. Following this, the `elm.json` needs to be manually re-attached.

It'd be great if I didn't have to go through this manual process on every start. Perhaps an option to attempt to automatically "auto discover" would do the job, or perhaps you have other ideas about what might be the best way to remove this friction.

Thanks very much for this fantastic tooling.

adamnfish:
After writing this it occurred to me to set the toochain to the bare command names, rather than qualified paths to the executables.

This works, but the `You must specify a path to the Elm compiler   [setup toolchain]` banner stays visible when the commands are configured this way. Perhaps allowing this banner to disappear when bare command names are included would resolve this issue more simply?

![intellij-elm](https://user-images.githubusercontent.com/29761/114216244-90306b80-995e-11eb-91a0-13d3bf7fb2ff.png)


Thanks again.

PR #751, opened 2021-04-30
Check bare elm executable

adamnfish:
Fixes https://github.com/klazuka/intellij-elm/issues/746

Tools like `fnm` manipulate the PATH to ensure executables are available, but they do not guarantee that the full path to those executables won't change. This means that after every restart, I have to re-configure the Elm toolchain from scratch.

This is easily avoided by specifying the bare executable names in the configuration, but this creates a new problem. The current implementation of `looksLikeValidToolchain` assumes that the configured path points to an executable file, and does not do PATH resolution.
https://github.com/klazuka/intellij-elm/blob/eafbfe8ffa8d2758de6a4da8c7233af02d14babd/src/main/kotlin/org/elm/workspace/ElmToolchain.kt#L34-L35

This PR updates the `looksLikeValidToolchain` function so that it allows the configuration to be provided "bare" (i.e. just the word `elm`). In this case it resolves the PATH to confirm that `elm` is available.

Practically speaking, this means that the "You must specify a path to the Elm compiler" banner won't show up all the time for these users.

I do not know how to actually use a locally-developed plugin in IntelliJ so I haven't seen this work "in the wild", but I have added tests for this functionality. These tests also cover the previously untested behaviour that confirms the configured compiler location is an executable file.

This is my first time writing Kotlin and my first time looking at an IntelliJ plugin. I'll be very happy to discuss the problem and this implementation, we might consider this a prompt rather than a suggestion!

A simpler implementation that allows a bare `elm` without verifying it is on the path can be found on another branch, in case that alternative seems like a better approach:
https://github.com/klazuka/intellij-elm/compare/master...adamnfish:allow-bare-elm-executable

Thank you for creating and maintaining this wonderful plugin!

### DIFF
--- a/src/main/kotlin/org/elm/ide/notifications/ElmNeedsConfigNotificationProvider.kt
+++ b/src/main/kotlin/org/elm/ide/notifications/ElmNeedsConfigNotificationProvider.kt
@@ -61,7 +61,7 @@ class ElmNeedsConfigNotificationProvider(
             return null
 
         val toolchain = project.elmToolchain
-        if (!toolchain.looksLikeValidToolchain(project)) {
+        if (!toolchain.looksLikeValidToolchain()) {
             return badToolchainPanel("You must specify a path to the Elm compiler")
         }
 
--- a/src/main/kotlin/org/elm/ide/notifications/ElmNeedsConfigNotificationProvider.kt
+++ b/src/main/kotlin/org/elm/ide/notifications/ElmNeedsConfigNotificationProvider.kt
@@ -61,7 +61,7 @@ class ElmNeedsConfigNotificationProvider(
             return null
 
         val toolchain = project.elmToolchain
-        if (!toolchain.looksLikeValidToolchain()) {
+        if (!toolchain.looksLikeValidToolchain(project)) {
             return badToolchainPanel("You must specify a path to the Elm compiler")
         }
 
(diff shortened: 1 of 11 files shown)


cies:
LGTM! I'm taking over the stewardship of this plugin and will merge your branch for the next release. Thanks!

adamnfish:
Wonderful news, thank you! 🙏

cies:
@adamnfish Could you merge `master` into that branch? I fixed the CI :)

adamnfish:
That's done, please let me know if I can do more to help.

cies:
@adamnfish I've merged your branch in 5.0.1...  Soon there will be a release on Github, and if its workable, i put it in the Marketplace.

>  please let me know if I can do more to help.

What about picking up the tools from the `$PATH` by default? E.g. when you specify no path(s)-to-tool(s) it simply tries to find them from the `$PATH`. Implementing this will require some UI changes (for which a mock-up would be great -- maybe we can find another plugin that already does this so we can copy their UX if that works nice)

Just an idea...

Other ways to help: the test suite needs some love, there are many tests currently "flaky" and the `intellij-rust` plugin (that this plugin is inspired by) has bits of code to fix the flakiness. Working on a codebase with a large non-flaky test suite is just so much better.

Good news: the tests now run (and pass for 99.5%+ except for the flaky ones) on `master`!

In case you did not know, this plugin has a channel on Elm Slack:

https://elmlang.slack.com/app_redirect?channel=intellij-plugin

PR #753, opened 2021-05-15
Add spacing around parens in tuples

jfmengels:
I am used to generating the type, then saving and seeing the type change because `elm-format` re-formats the type annotation.

This change makes the type look more like what elm-format would format it to. Records also have this spacing so this would be consistent with that too.


cies:
LGTM!

cies:
@jfmengels Could you merge `master` into that branch? I fixed the CI :) Also that will likely fix the conflict.

PR #754, opened 2021-05-20
Support parsing empty records in patterns

jfmengels:
## Issue

The following valid Elm code is not being properly parsed

```elm
emptyRecordFn {} = 1
```

## Pull request status

This pull request does two things:
- Add failing test cases (in the `complete` suite)
- Change the parser to allow an optional list of valid fields 

## Note

Using the "Add type annotation" intention for `a {} = 1` now generates `a : { a |  } -> number` as a type annotation. The compiler itself has a similar problem, which I have pointed at in https://github.com/elm/compiler/issues/2204. I think this is an okay problem to have, especially since there isn't any valid type that could be infered here.


cies:
LGTM! I'm taking over steward ship of this project and will merge this for the next release.

cies:
@jfmengels Could you merge `master` into that branch? I fixed the CI :)

PR #764, opened 2021-07-28
WIP: Remove bintray repository URLs and update dependencies

arian:
I tried to look into #763, but couldn't build the project anymore, because bintray URLs don't work anymore.

```
* What went wrong:
A problem occurred configuring root project 'intellij-elm'.
> Failed to notify project evaluation listener.
   > Could not resolve all files for configuration ':compileOnly'.
      > Could not resolve org.jetbrains.intellij.deps.jflex:jflex:1.7.0-1.
        Required by:
            project :
         > Could not resolve org.jetbrains.intellij.deps.jflex:jflex:1.7.0-1.
            > Could not get resource 'https://dl.bintray.com/kotlin/kotlin-dev/org/jetbrains/intellij/deps/jflex/jflex/1.7.0-1/jflex-1.7.0-1.pom'.
               > Could not HEAD 'https://dl.bintray.com/kotlin/kotlin-dev/org/jetbrains/intellij/deps/jflex/jflex/1.7.0-1/jflex-1.7.0-1.pom'. Received status code 403 from server: Forbidden
         > Could not resolve org.jetbrains.intellij.deps.jflex:jflex:1.7.0-1.
            > Could not get resource 'http://dl.bintray.com/jetbrains/markdown/org/jetbrains/intellij/deps/jflex/jflex/1.7.0-1/jflex-1.7.0-1.pom'.
               > Could not HEAD 'http://dl.bintray.com/jetbrains/markdown/org/jetbrains/intellij/deps/jflex/jflex/1.7.0-1/jflex-1.7.0-1.pom'. Received status code 403 from server: Forbidden
         > Could not resolve org.jetbrains.intellij.deps.jflex:jflex:1.7.0-1.
            > Could not get resource 'https://jetbrains.bintray.com/intellij-third-party-dependencies/org/jetbrains/intellij/deps/jflex/jflex/1.7.0-1/jflex-1.7.0-1.pom'.
               > Could not HEAD 'https://jetbrains.bintray.com/intellij-third-party-dependencies/org/jetbrains/intellij/deps/jflex/jflex/1.7.0-1/jflex-1.7.0-1.pom'. Received status code 403 from server: Forbidden
```

So i tried fixing that:

- Bintray has Sunset
- upgrade grammarkit to 2021.1.3
- upgrade idea gradle plugin to 1.1
- upgrade kotlin to 1.5
- Update gradle to 6.9

That fixes the compile errors, but at the moment i'm still running into:

```
> Task :instrumentCode FAILED
[...]

PR #765, opened 2021-07-28
Change until-build to 212 to support IntelliJ 2021.2

yonigibbs:
This is a lower-tech solution than #764, in case that PR is deemed too risky. This one simply increments the `until-build` number from 211 to 212. I can't currently test whether this builds successfully as the code won't compile for me locally for reasons outlined in #764. But I'm submitting this anyway in the hope that maybe it'll pass the CI/CD pipeline if all the artefacts are already cached somewhere on the build system. Whether this will pass the CI build remains to be seen, given the code still uses jcenter/bintray, which are now retired. If it does pass, this might be an easier fix to get the plugin working with IntelliJ 2021.2. Though the larger-scale upgrade of things in #764 is still something that should be done at some point.

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -331,7 +331,7 @@
     Note that the range is half-open: [since-build, until-build), but if you use a wildcard in `until`,
     it will be inclusive (i.e. 203.* is inclusive for 2020.3.x)
      -->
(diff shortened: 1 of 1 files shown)


yonigibbs:
Nope, didn't work unfortunately. Strangely, the build passed on Windows, so the artefacts must have been cached somewhere I think. But on Ubuntu they failed with the same errors described in #764. Am closing this PR.

Issue #776, opened 2021-12-01
Deprecation warning for 'getId'


karvanit:
Running the Elm plugin in Intellij Ultimate 21.3.2 gives the following warning:

```
com.intellij.diagnostic.PluginException: The default implementation of method 'getId' is deprecated, you need to override it in 'class org.elm.ide.test.run.ElmTestConfigurationFactory'. The default implementation delegates to 'getName' which may be localized, but return value of this method must not depend on current localization. [Plugin: org.elm.klazuka]
	at com.intellij.ide.plugins.PluginManagerCore.createPluginException(PluginManagerCore.java:270)
	at com.intellij.diagnostic.PluginProblemReporterImpl.createPluginExceptionByClass(PluginProblemReporterImpl.java:12)
	at com.intellij.diagnostic.PluginException.createByClass(PluginException.java:83)
	at com.intellij.diagnostic.PluginException.reportDeprecatedDefault(PluginException.java:110)
	at com.intellij.execution.configurations.ConfigurationFactory.getId(ConfigurationFactory.java:75)
	at com.intellij.execution.impl.RunManagerImpl.getFactory(RunManagerImpl.kt:1029)
	at com.intellij.execution.impl.RunManagerImpl.getFactory(RunManagerImpl.kt:1012)
	at com.intellij.execution.impl.RunnerAndConfigurationSettingsImpl.readExternal(RunnerAndConfigurationSettingsImpl.kt:204)
	at com.intellij.execution.impl.RunConfigurationSchemeManager.readData(RunConfigurationSchemeManager.kt:60)
	at com.intellij.execution.impl.RunConfigurationSchemeManager.createScheme(RunConfigurationSchemeManager.kt:43)
	at com.intellij.execution.impl.RunConfigurationSchemeManager.createScheme(RunConfigurationSchemeManager.kt:21)
	at com.intellij.configurationStore.LazySchemeProcessor.createScheme$default(scheme-impl.kt:64)
	at com.intellij.configurationStore.schemeManager.SchemeLoader.loadScheme(schemeLoader.kt:177)
	at com.intellij.configurationStore.schemeManager.SchemeManagerImpl$loadSchemes$isLoadOnlyFromProvider$2.invoke(SchemeManagerImpl.kt:215)
	at com.intellij.configurationStore.schemeManager.SchemeManagerImpl$loadSchemes$isLoadOnlyFromProvider$2.invoke(SchemeManagerImpl.kt:42)
[...]

PR #781, opened 2022-02-16
Fix error about ElmTestConfigurationFactory.getId()

karvanit:
Use hardcoded string instead of falling back on getName().
Closes issue #776.

### DIFF
--- a/src/main/kotlin/org/elm/ide/test/run/ElmTestConfigurationFactory.kt
+++ b/src/main/kotlin/org/elm/ide/test/run/ElmTestConfigurationFactory.kt
@@ -10,6 +10,8 @@ class ElmTestConfigurationFactory internal constructor(type: ConfigurationType) 
     override fun createTemplateConfiguration(project: Project) =
             ElmTestRunConfiguration(project, this, "Elm Test")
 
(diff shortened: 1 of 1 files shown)


janekx21:
Can this please be approved? @karvanit @FloWi

klazuka:
Unfortunately due to the bintray issue, the project no longer builds. And sadly it requires updating a bunch of things like GrammarKit and IntelliJ, but that causes a bunch of failures. Someone would need to continue the work in #764 before we could do a new release.

clojj:
> Unfortunately due to the bintray issue, the project no longer builds. And sadly it requires updating a bunch of things like GrammarKit and IntelliJ, but that causes a bunch of failures. Someone would need to continue the work in #764 before we could do a new release.

@klazuka 
Hi Keith !
from the current IJ plugin platform I managed to renovate several things, fixed some deprecations, improved compiler toolwindow and integrated elm-review 
would you be interested in a PR ?
The missing thing would be to finalize the new publishing of the plugin, should be straightforward since it is using latest IJ gradle build.

klazuka:
@clojj yeah, post the PR, that sounds good.

PR #783, opened 2022-02-23
Version 5.0.0 proposal

clojj:
Hi, a lot of things.. but it has already been tested by a few people (other than myself)

- plugin-build updated: https://github.com/JetBrains/intellij-platform-plugin-template served as basis (I choose not to fork from there though, since we'd loose history)
- initial support for Lamdera projects (which have no usual main entrypoint, but have 2 fixed app functions for backend and frontend)
- much improved implementation of compiler & review toolwindows and related message-panels (errors are displayed in standard error-view, navigation is much better, IJ color appearance is honoured now, that is light, dark, ...), summed up in "much less code" with better quality
- initial support for elm-review
- some deprecations fixed
- general Bugfix for ElmBuildAction "ERROR - ution.process.OSProcessHandler - Synchronous execution on EDT"
- dependencies updated, last but not least Kotlin 1.6

What is *missing* is the part in the README which was merged from the template:
-> ## Template ToDo list
which amounts to *setting up publishing* and activate the workflows in the 'commented out' directory `.github_`

Also, I din't  use qodana yet.
Some unit-tests are still being worked on/created, but besides some odd phenomena of needing to run some tests twice before green, the unit-tests run Ok. With Github workflows in place, this has probably be to checked anyway in more detail.

What has been tested many times already with this version, is just generating a ZIP distribution, which was confirmed installable in IntelliJ Idea and WebStorm

Best,
Joerg


clojj:
Because of an unknown Git-related problem in the fork, this PR will have to be re-created from a new fork.

PR #784, opened 2022-02-25
Enable ElmIncompletePatternInspection inspection on the of keyword

jfmengels:
Very often I (and others) write a case expression like `case x of` and the plugin is able to infer that there are missing cases and which ones (or that you need a `_` branch).

When I finish to write the `of` keyword, to make use of the inspection and automatic fixes, I have to go back to the `case` keyword. Instead, I'd like to have fixes from the `of` keyword where my editor cursor is at that moment. The `of` keyword is also more accessible by mouse because you can click anywhere on the line after the `of` keyword, but having to go to `case` requires more precision.

This PR makes the inspections available on `case` available also on `of`.

### DIFF
--- a/src/main/kotlin/org/elm/ide/inspections/ElmIncompletePatternInspection.kt
+++ b/src/main/kotlin/org/elm/ide/inspections/ElmIncompletePatternInspection.kt
@@ -7,6 +7,9 @@ import com.intellij.psi.PsiElement
 import org.elm.ide.inspections.MissingCaseBranchAdder.Result.MissingVariants
 import org.elm.ide.inspections.MissingCaseBranchAdder.Result.NoMissing
 import org.elm.lang.core.psi.ElmPsiElement
+import org.elm.lang.core.psi.ElmTypes
+import org.elm.lang.core.psi.directChildren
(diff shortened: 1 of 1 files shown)


pravdomil:
see https://github.com/klazuka/intellij-elm/pull/757

cies:
I'll merge this. Nice improvement!

jfmengels:
One downside of this if I'm not mistaken, is that this duplicates the incomplete pattern error in the errors panel. Another possibility is to put the underline on the entire `case ... of` section (possibly spanning a lot of code).

If we could find a way to not duplicate the error, that would be nice :man_shrugging:

cies:
@jfmengels  Could you merge `master` into that branch? I fixed the CI :)

cies:
@jfmengels I've tried to merge this in the `5.0.1` branch (which itself is a branch off of `master`) and there are a bunch of tests that fail because of it, namely:

image

I have tried to fix them, but it started to take more time than I had available for it... I kind of hope you could take a look at it. :)

Best is to first merge master into this your branch, as most  tests on master currently pass (a few I did not manage to fix yet and quite a few are "flaky").

Issue #4, opened 2017-12-16
Improve Find Usages by classifying the usage by type


klazuka:
Right now when you do Find Usages, everything shows up as "unclassified usage". We need to implement the `UsageTypeProvider` extension point to provide this capability.

klazuka:
This looks easy to implement (see intellij-rust `org.rust.ide.search.RsUsageTypeProvider` and `org.rust.ide.search.RsFindUsagesTest`).

The bigger task is probably just figuring out which categories would be useful to distinguish between.

PR #802, opened 2024-04-03
Add ElmUsageTypeProvider by @ktosiek

cies:
Add code I found on a branch by @ktosiek implementing an `ElmUsageTypeProvider`


cies:
Maybe fixes #4

BenjaminDebeerst:
This PR, as it stands does not fix #4 - the usages are still all "Unclassified".

The `getUsageType` function always ends up in the [default case returning null.](https://github.com/intellij-elm/intellij-elm/pull/802/files#diff-435d9b90e8808a0ead6684e2bf9ee3a4c051a75501ad9e0569007f12c894d4e2R24).

I'm looking into it.

ktosiek:
@ktosiek here: it seems what I've implemented only works when you ctrl-click *and* send those results to the search window. I hoped it will work in the ctrl-click/ctrl-b popup - but it didn't, I've never found why.

BenjaminDebeerst:
Interesting. For me, the `UsageTypeProvider` adds a grouping to the results of the find usages dialogue (Show Usages / Ctrl-Alt-7 or Find Usages / Ctrl-Alt-7). That is, how to group search results and what the display string becomes for what currently is "Unclassified" in this screenshot:

Screenshot from 2024-04-05 23-25-26

I personally don't think this is critical functionality and don't use it much, but I found it a good first issue to learn about the plugin :-)

And I made some progress already, in essence it is only about adding more branches to the `getUsageType` function to cover all PsiElements that could be referencing the element for which usages were searched.

I don't know how this relates to the Ctrl-b popup, which I've never used before. I also don't exactly understand what you mean by "when you ctrl-click and send those results to the search window" - how would I do that?

BenjaminDebeerst:
And yes indeed, the ctrl-click/ctrl-b popup you're referring to, the "Show Usages" dialogue, does not have this grouping. Or only partially - it does allow some filtering ("Show read access", "Show write access" etc), which I'd expect to be informed by this implementation as well.

ktosiek:
@BenjaminDebeerst I see you've found what I was talking about, thank you for describing it better than I did ^_^'

I think I don't have anything to add at this point, you seem to have at least as much understanding of the problem as I had when commiting this code.

cies:
Merged master into this so we can see CI results.

I'm not really sure based on the discussion above: should I still merge this into the next release?

PR #803, opened 2024-04-03
Improve ElmIncompletePatternInspection (@jfmengels)

cies:
Found this on @jfmengels' Github, looks legit :)


lydell:
Oh, this moves the underline from `case` to `of`? That would be amazing!

jfmengels:
This is a duplicate of https://github.com/intellij-elm/intellij-elm/pull/784 by the way, I had already made a PR for this.

@lydell it doesn't move the underline, it duplicates it. That way you can use it on both `case` and `of`.

cies:
@jfmengels Oops missed it. Closing this.

PR #804, opened 2024-04-03
Fix single union pattern match

### LINKED ISSUE #742: Parse error when pattern matching in function-argument
Tested in latest release, 4.3.1

The code below is valid elm-code, but `f2` yields a syntax error.
It seems like itellij-elm does not like pattern matches in function without parens

A workaround is to wrap every such argument in parens. This is however not possible if using elm-format, since it automatically removes the parens unless the type constructor in question accepts one or more arguments.

view in [ellie-app](https://ellie-app.com/cFymb5YVGzba1)

```elm
module Main exposing (..)

import Html


type MyType
    = TheOnlyConstructor

-- valid syntax according to intellij-elm
f1 : MyType -> String
f1 (TheOnlyConstructor) =
    "f1 output"


-- invalid syntax according to intellij-elm
f2 : MyType -> String
f2 TheOnlyConstructor =
    "f2 output"


main =
    Html.div []
        [ Html.div [] [ Html.text <| f1 TheOnlyConstructor]
        , Html.div [] [ Html.text <| f2 TheOnlyConstructor]
        ]

```

Error message:

image

cies:
Closes #742


cies:
Wrong branch. Will recreate the PR with the correct branch.

PR #806, opened 2024-04-04
Remove Lamdera notes that are no longer necessary

supermario:
The latest version of Lamdera uses `~/.elm/0.19.1` again, and all linux builds are now statically built so shouldn't have any external lib dep issues anymore.

One thing I'm unsure about is whether we need to advise the user to set their elm path to the lamdera binary path? Or will intellij-elm automatically figure that out?

### DIFF
--- a/docs/features/lamdera.md
+++ b/docs/features/lamdera.md
@@ -8,20 +8,3 @@ A Lamdera project currently consists of:
 
 To get going, you can download Lamdera's compiler here:
 [Download lamdera executable](https://dashboard.lamdera.app/docs/download)
-
-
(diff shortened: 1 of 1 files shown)


cies:
@supermario thanks, I'll merge this.

> One thing I'm unsure about is whether we need to advise the user to set their elm path to the lamdera binary path? Or will intellij-elm automatically figure that out?

I'm not using Lamdera (I love your presentation about it on youtube in which you slash boilerplate), but in the places I use Elm we have backend tech already in place. Maybe you can figure this out by doing some test.

PR #808, opened 2024-04-06
Extend FindUsagesProvider and UsageTypeProvider

### LINKED ISSUE #4: Improve Find Usages by classifying the usage by type

BenjaminDebeerst:
Here's my attempt in making the UsageTypeProvider more useful.

* Provide descriptive string for all `PsiElement`s for which usages can be searched (that is, all implementations of `ElmNamedElement`)
* Exhaustively categorize all usage types
* Simplify implementation to implement `UsageTypeProvider` instead of `UsageTypeProviderEx`: The latter allows to distinguish the usage type depending on the target element of the usage, which this implementation does not use

The usage type categorization should be exhaustive for all possible results of a usage search - if a usage type is ever categorized as "Other" or "Unknown", this is something that can easily be improved.

Please give this a shot by searching for usages in various places, feedback on any aspect is very welcome.

This is entirely manually tested so far. The unit tests for finding usages may not be green with this change, I didn't get the tests to run so far an won't be able to spend time making them run today :-)

~~(Note that the base branch of this PR is set to the branch for #802, which this builds on top of.)~~
Now merges into master.

Fixes #4.


cies:
I've merged master into #802 (branch `add-ElmUsageTypeProvider`). If you can merge that branch into this that'd be great (fixes CI)

PR #833, opened 2025-02-01
improved syntax highlight

ilsasdo:
With this PR I'm trying to improve the Syntax Highlight for ELM.

Current highlight support is very minimal and doesn't make any difference for various elements like: function calls of different scopes and arguments with some hard to read results.

Since in Elm basically everything is a Function Call or Arguments, it's important to know at least the _scope_ of the function being called.

For this reason, I introduced different level of highlighting based on scope:
- Imported Module, Imported Module Function
- Ports
- Function Calls related to the current file (and their arguments)
- Function Calls declared in `let-in` blocks (and their arguments)
- Anonymous Function Calls (and their arguments)
- Pattern match arguments.

Here two comparison images:
1. previous this PR: 
previous

2. after this PR (default color scheme): 
after

3. with some random colors: 
after_custom_colors


clojj:
@ilsasdo 
This is excellent! and thx for these detailed shots

I will give it a try, if I manage to get releases rolling again (big If at the moment though)
Btw: did you build a release locally for your own use

